PR #1, opened 2022-06-23
Set nodePort for Service

pbchekin:
Use `api.service.nodePort` to set `nodePort` for a NodePort Service.


gabcoyne (approved):
This looks good

PR #4, opened 2022-06-24
Update Readme.md

khuyentran1401:
Update readme

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -135,4 +135,4 @@ cd prefect-helm
 
 ## Options:
 
-See comments in `values.yaml`.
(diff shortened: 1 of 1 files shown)


Readme.md
```diff
@@ -7,7 +7,7 @@
 $ helm repo add prefect https://prefecthq.github.io/prefect-helm
```
gq1, line 7: the repo name not math, prefect of prefecthq?

khuyentran1401: You mean line 7? `$ helm repo add prefect https://prefecthq.github.io/prefect-helm` 

I kept that line the same if that's what you refer to

gq1: if you don't change line 7, and 8, and then line 9 needs to be:

`helm install my-release prefect/<chart>`

PR #11, opened 2022-07-26
Fix oversights regarding namespace and ports in templates

clamydo:
I've fixes some minor oversights when setting the namespace and the port in the chart templates. Parameterised them - I hope I've done it correctly.

### DIFF
--- a/charts/prefect-orion/templates/api/deployment.yaml
+++ b/charts/prefect-orion/templates/api/deployment.yaml
@@ -36,7 +36,7 @@ spec:
           imagePullPolicy: {{ .Values.api.image.pullPolicy }}
           command: ["prefect", "orion", "start", "--host", "0.0.0.0", "--log-level", "WARNING"]
           ports:
-          - containerPort: 4200
+          - containerPort: {{ .Values.api.service.port }}
           resources:
             {{- toYaml .Values.api.resources | nindent 12 }}
           env:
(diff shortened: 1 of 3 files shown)


charts/prefect-orion/templates/api/deployment.yaml
```diff
@@ -36,7 +36,7 @@ spec:
           imagePullPolicy: {{ .Values.api.image.pullPolicy }}
           command: ["prefect", "orion", "start", "--host", "0.0.0.0", "--log-level", "WARNING"]
           ports:
-          - containerPort: 4200
+          - containerPort: {{ .Values.api.service.port }}
```
jamiedick, line 39: I dont see the api service port in the values file -- can you add in and perhaps set the default to 4200?

jamiedick (changes requested):
minor change requested - otherwise awesome work!

charts/prefect-orion/templates/api/deployment.yaml
clamydo: Hey, glad to help. 

Isn't it? https://github.com/PrefectHQ/prefect-helm/blob/26dca12e31fae629bb3d4ffc0af1bbdc0cffbb77/charts/prefect-orion/values.yaml#L7

jamiedick (approved):
🚀

PR #13, opened 2022-07-28
Fix display the prefect API URL in agents notes

clamydo:
Previously a non-existing key was referenced. I suppose `agent.prefectApiUrl` is the correct key.


jamiedick (approved):
nice - good catch!

PR #16, opened 2022-07-29
Adapt env-unwrap function for orion

clamydo:
Adapting the `env-unwrap` function to make sense for orion, i.e., setting `PREFECT_ORION_` env vars. I guess, this was a left-over from prefect-server?


jamiedick (approved):
Thanks @clamydo for the rapid feedback!

PR #18, opened 2022-07-29
Release version

gabcoyne:
Release candidate


charts/prefect-agent/Chart.yaml
```diff
@@ -3,7 +3,7 @@ name: prefect-agent
 description: Prefect orion application bundle
 type: application
 home: https://github.com/PrefectHQ
-version: 2022.6.30
+version: 2022.7.29
```
jamiedick, line 6: can we adopt some semvar instead of date?

jamiedick:
also, should we bump the prefect version to 2.0 instead of the ref to beta?

jamiedick (approved):
🚀

PR #27, opened 2022-09-08
Charts enhancement

john-jam:
This PR tries to fix some issues I had while testing the helm charts in this repository:
- Remove the unused `prefectVersionTag` and set `2-latest` as the default image for `prefect-orion` instead of `latest-python3.9` that was prefect 1.0 based.
- Define `apiKeySecret` as optional with default empty values to avoid creating a fake secret if the agent is not connected to prefect cloud.
- Fix the `existingSecret` configuration for postgres connection string. It is now possible to define the connection string with our password in our own secret.
- Rename the orion `debug_enabled` to `debugEnabled` to match the `agent` config

Please let me know if there is anything I should change and thank you for this project!


jamiezieziula (approved):

john-jam:
Do you know when the next release is planned for those 2 charts ?

jamiezieziula:
@john-jam - its out now!

PR #30, opened 2022-09-13
[prefect-orion] hpa looking for values on wrong places

bosmak:
The values.yaml indicates that the configuration for the hpa would be under `api.autoscalling.X` but the current file looks at `autoscalling.X`


jamiezieziula (approved):
lgtm - thanks! 🚀

PR #32, opened 2022-09-14
[prefect-orion] create service account

bosmak:
Currently, the deployment sets a service account to the deployment but it isn't created anywhere.


jamiezieziula:
this is a cool reference - https://github.com/kubernetes-sigs/descheduler/blob/master/strategies_diagram.png

jamiezieziula (approved):
awesome - this looks good to me!

PR #33, opened 2022-09-14
[prefect-orion] Add values that can be used but aren't on the default values.yaml or README

bosmak:
I made the `api.image.pullSecrets` work the same way across both charts.

Besides that mostly is documentation.


jamiezieziula (approved):
nice 🚀

PR #38, opened 2022-09-16
[orion] Fix pathing reference to postgresql.auth.existingSecret

### LINKED ISSUE #35: `existingSecret` config path documentation is wrong
Documented path is `postgresql.auth.existingSecret`

[Path used in source code](https://github.com/PrefectHQ/prefect-helm/blob/f0e838aa20402fa8c23cb4fca8e6c0576d288b1d/charts/prefect-orion/templates/_helpers.tpl#L136) is `postgresql.existingSecret`

Either doc or sources should be fixed.

Because of that issue, it currently takes a while to figure out why external db authentication does not work.

jamiezieziula:
Closes #35


jawnsy (approved):
LGTM

PR #40, opened 2022-09-16
[orion] Add prefectConfig to values file

### LINKED ISSUE #15: Information about `prefectConfig` is missing in the README for `prefect-orion`
Setting `prefectConfig` is useful to modify prefect's environment variables. However, a documentation, how to use it is missing from the README. 

The Chart README seems to be semi-generated. What is a good place to document this?

jamiezieziula:
Closes #15

### DIFF
--- a/charts/prefect-orion/values.yaml
+++ b/charts/prefect-orion/values.yaml
@@ -133,7 +133,7 @@ postgresql:
 # Values placed here will result in environment variable definitions
 # starting with "PREFECT_ORION_" and will be passed to relevant
 # services
-prefectConfig:
+prefectConfig: {}
 
 serviceAccount:
   # Specifies whether a service account should be created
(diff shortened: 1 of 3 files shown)


charts/prefect-orion/values.yaml
```diff
@@ -130,6 +130,11 @@ postgresql:
     # -- Version tag, corresponds to tags at https://hub.docker.com/r/bitnami/postgresql/
     tag: 14.3.0
 
+# Values placed here will result in environment variable definitions
+# starting with "PREFECT_ORION_" and will be passed to relevant
+# services
+prefectConfig:
```
jawnsy, line 136:
```suggestion
prefectConfig: {}
```

if you use an empty object as a default, then it simplifies some of the logic - similar to what we're doing for annotations

the type info shown by helm-docs also seems weird (string instead of object?)

```
| prefectConfig | string | `nil` |  |
| serviceAccount.annotations | object | `{}` |  |
```

jawnsy (dismissed):

jawnsy (approved):

PR #43, opened 2022-09-18
fix: use prefect agent start -q

tekumara:
to silence this startup message:
```
Agents now support multiple work queues. Instead of passing a single argument, 
provide work queue names with the `-q` or `--work-queue` flag: `prefect agent 
start -q kubernetes`
```


jamiezieziula (approved):
nice - lgtm!

PR #48, opened 2022-09-21
[orion] chart overhaul

### LINKED ISSUE #44: Add configuration settings inspired by Bitnami's chart templates
Bitnami charts include a number of customization capabilities that may be useful to adopt in this chart as well. They provide a template with some comprehensive values here: https://github.com/bitnami/charts/blob/master/template/CHART_NAME/values.yaml

Open questions:

* Which of these features would be useful to replicate in Prefect's Helm charts?
* Should we use the same value names, or choose different ones?
* How should we handle backward compatibility for users of our existing charts? Or do we make this a breaking change?

jamiezieziula:
This PR rebuilds the prefect-orion chart using bitnamis [common](https://github.com/bitnami/charts/tree/master/template/CHART_NAME) as its base.

Closes https://github.com/PrefectHQ/prefect-helm/issues/44


charts/prefect-orion/templates/service.yaml
```diff
@@ -1,16 +1,38 @@
...
+  type: {{ .Values.service.type }}
+  {{- if and .Values.service.clusterIP (eq .Values.service.type "ClusterIP") }}
+  clusterIP: {{ .Values.service.clusterIP }}
+  {{- end }}
+  {{- if eq .Values.service.type "NodePort" }}
+  externalTrafficPolicy: {{ .Values.service.externalTrafficPolicy | quote }}
+  {{- end }}
   ports:
-    - port: {{ .Values.api.service.port }}
+    - name: orion # will this be fine?
```
jamiezieziula, line 29: open to recommendations for port name ehre

jimid27 (dismissed):

charts/prefect-orion/templates/service.yaml
parkedwards: could just be very literal with it, something like `orion-svc-port`

charts/prefect-orion/Chart.yaml
```diff
@@ -1,21 +1,27 @@
 apiVersion: v2
-name: prefect-orion
+appVersion: latest
+dependencies:
+  - name: common
+    repository: https://charts.bitnami.com/bitnami
+    tags:
+      - bitnami-common
+    version: 2.0.2
```
parkedwards, line 8: one thing to keep in mind about switching to a bitnami common base is that we'll have to keep this relatively updated to latest, since they started having to drop anything older than 6 months prior from their root manifest bc their cdn started timing out

https://github.com/bitnami/charts/issues/10539

parkedwards (dismissed):
🚀

charts/prefect-orion/Chart.yaml
jamiezieziula: hm im not sure what you're suggesting @madkinsz

jamiezieziula: good to know - I'll add an item to determine a good way to track upstream dependencies & releases

parkedwards (approved):

PR #49, opened 2022-09-21
[orion] Add support for running unprivileged

jamiezieziula:
Run as a non-root user with a read-only filesystem, mounting an emptyDir volume for temporary state files. This also prevents privilege escalation by default.


jawnsy (dismissed):

charts/prefect-orion/templates/service.yaml
```diff
@@ -26,7 +26,7 @@ spec:
   externalTrafficPolicy: {{ .Values.service.externalTrafficPolicy | quote }}
   {{- end }}
   ports:
-    - name: orion # will this be fine?
+    - name: orion-svc-port
```
jawnsy, line 29: there's also an appProtocol field, which is useful to set to 'http' or 'https' depending what we have here https://kubernetes.io/docs/concepts/services-networking/service/#application-protocol

jawnsy (approved):
lgtm

PR #51, opened 2022-09-22
Update versioning & release workflow

jamiezieziula:
This PR updates the way we tag chart versions as well as reference app version tags in values.yaml

### DIFF
--- a/.github/workflows/helm-release.yaml
+++ b/.github/workflows/helm-release.yaml
@@ -81,7 +81,7 @@ jobs:
         run: |
           cd charts
           # Update the prefect version tag
-          sed -i -E "\s{0}2-latest \"$PREFECT_VERSION\"/" prefect-orion/values.yaml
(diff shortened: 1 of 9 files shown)


jawnsy (approved):

.github/workflows/helm-release.yaml
```diff
@@ -20,7 +20,12 @@ jobs:
           # Enable pipefail so git command failures do not result in null versions downstream
           set -x
 
-          echo ::set-output name=VERSION::$(echo $GITHUB_REF | cut -d / -f 3)
+          echo ::set-output name=RELEASE_VERSION::$(echo $GITHUB_REF | cut -d / -f 3)
```
jawnsy, line 23: this sets a job output, but it's used as an environment variable later, so I'm not sure this actually works

you might have to write this to the $GITHUB_ENV file if you expect to use it as an env var

jamiezieziula: it works! - https://github.com/PrefectHQ/prefect-helm/actions/runs/3102476140/jobs/5024809170

Issue #57, opened 2022-09-22
`workQueueName` also needs a value for the agent to run


padbk:
Kind of related, but `workQueueName` also needs a value for the agent to run, due to the addition of the `-q` variable here https://github.com/PrefectHQ/prefect-helm/blob/8434d5e27ad6a996d7ebb36e9c75e3ee556f65fd/charts/prefect-agent/templates/deployment.yaml#L53

_Originally posted by @padbk in https://github.com/PrefectHQ/prefect-helm/issues/53#issuecomment-1255127774_

jamiezieziula:
Hi @padbk - a workqueue name is required either way.  Before we were defaulting to a value of `prd` but have since removed that.  We can revert to specifying a default value, but either way a name is needed.

PR #58, opened 2022-09-22
[agent] Adjust agent configuration settings

jamiezieziula:
Related to #57, this better structures what configuration is needed for all agent deployments and delineates what is needed for agent deployments with prefect cloud versus configuration for agent deployments with a local orion instance.

### DIFF
--- a/charts/prefect-agent/README.md
+++ b/charts/prefect-agent/README.md
@@ -34,7 +34,7 @@ Prefect Agent application bundle
 | agent.cloudApiConfig.apiKeySecret.name | string | `"prefect-api-key"` | prefect API secret name |
 | agent.cloudApiConfig.cloudUrl | string | `"https://api.prefect.cloud/api"` | prefect cloud API url; the full URL is constructed as https://cloudUrl/accounts/accountId/workspaces/workspaceId |
 | agent.cloudApiConfig.workspaceId | string | `""` | prefect workspace ID |
-| agent.config.workQueueNames | list | `["default"]` | names of prefect workqueues the agent will poll |
+| agent.config.workQueues | list | `["default"]` | names of prefect workqueues the agent will poll |
 | agent.containerSecurityContext.allowPrivilegeEscalation | bool | `false` | set agent containers' security context allowPrivilegeEscalation |
 | agent.containerSecurityContext.readOnlyRootFilesystem | bool | `true` | set agent containers' security context readOnlyRootFilesystem |
 | agent.containerSecurityContext.runAsNonRoot | bool | `true` | set agent containers' security context runAsNonRoot |
--- a/charts/prefect-agent/README.md
+++ b/charts/prefect-agent/README.md
@@ -28,13 +28,13 @@ Prefect Agent application bundle
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | agent.affinity | object | `{}` | affinity for agent pods assignment |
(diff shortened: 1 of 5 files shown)


charts/prefect-agent/README.md
```diff
@@ -28,13 +28,13 @@ Prefect Agent application bundle
...
-| agent.config.apiUrl | string | `"https://api.prefect.cloud"` | prefect API url |
-| agent.config.workQueueName | string | `""` | prefect workqueue name |
-| agent.config.workspaceId | string | `""` | prefect workspace ID |
+| agent.apiConfig | string | `"cloud"` | one of 'cloud' or 'orion' |
+| agent.cloudApiConfig.accountId | string | `""` | prefect account ID |
+| agent.cloudApiConfig.apiKeySecret.key | string | `"key"` | prefect API secret key |
+| agent.cloudApiConfig.apiKeySecret.name | string | `"prefect-api-key"` | prefect API secret name |
+| agent.cloudApiConfig.cloudUrl | string | `"https://api.prefect.cloud/api"` | prefect cloud API url; the full URL is constructed as https://cloudUrl/accounts/accountId/workspaces/workspaceId |
+| agent.cloudApiConfig.workspaceId | string | `""` | prefect workspace ID |
+| agent.config.workQueueName | string | `"default"` | prefect workqueue name |
```
jawnsy, line 37: if we're going to make breaking changes now, then I'd propose that we change this type to an array of strings (for multi work queue support) and also just call this `workQueue` or `workQueues`, because users can specify either a name or UUID here

something like:

```yaml
agent:
  config:
    workQueues:
      - abc
      - def
```

I think there's template functions that can do an array join for us

charts/prefect-agent/templates/_helpers.tpl
```diff
@@ -14,9 +14,9 @@ Create the name of the service account to use
     Define API URL for workspace or for
 */}}
 {{- define "agent.apiUrl" -}}
-{{- if and .Values.agent.config.accountId .Values.agent.config.workspaceId }}
-{{- printf "%s/api/accounts/%s/workspaces/%s" .Values.agent.config.apiUrl .Values.agent.config.accountId .Values.agent.config.workspaceId | quote }}
+{{- if eq .Values.agent.apiConfig "cloud" }}
+{{- printf "%s/accounts/%s/workspaces/%s" .Values.agent.cloudApiConfig.cloudUrl .Values.agent.cloudApiConfig.accountId .Values.agent.cloudApiConfig.workspaceId | quote }}
```
jawnsy, line 18: I'm wondering if we could handle this instead by showing the API URL in the UI somewhere (we could have a page that explains to people how to install an agent for a given work queue, with an example command to run in the case of the Helm chart)

so maybe orion/cloud don't really need to be different after all?

charts/prefect-agent/README.md
jamiezieziula: good call - done!

jawnsy (dismissed):

charts/prefect-agent/templates/_helpers.tpl
madkinsz: I don't think the Helm chart is mainstream enough to target from the UI. The core library already templates/generates the API URL when users do `prefect cloud login` and we've established that they should not be responsible for creating the path to their workspace. I think it's reasonable for the chart to do so as well 🤷‍♀️

gabcoyne (approved):

PR #60, opened 2022-09-23
pass `appVersion` to container labels

jamiezieziula:


charts/prefect-agent/templates/deployment.yaml
```diff
@@ -24,6 +24,7 @@ spec:
       {{- end }}
       labels: {{- include "common.labels.standard" . | nindent 8 }}
         app.kubernetes.io/component: agent
+        prefect-version: {{ .Chart.AppVersion }}
```
jawnsy, line 27: I guess this isn't included in the common labels?

jawnsy (approved):

charts/prefect-agent/templates/deployment.yaml
jamiezieziula: correct!

jamiezieziula: https://github.com/bitnami/charts/blob/master/bitnami/common/templates/_labels.tpl

PR #61, opened 2022-09-23
Fix a typo when checking the support for ingress classname in orion

jlanglois-jam:
When installing the `orion` chart with `ingress.className` defined, the following error occurs:
```
Error: template: prefect-orion/templates/ingress.yaml:23:51: executing "prefect-orion/templates/ingress.yaml" at <include "common.ingress.supportsIngressClassName" .>: error calling include: template: no template "common.ingress.supportsIngressClassName" associated with template "gotpl"
```

I think there's a typo in the helper call since it's a lower `n` in the `bitnami/common` documentation: https://artifacthub.io/packages/helm/bitnami/common/2.0.2#ingress

### DIFF
--- a/charts/prefect-orion/templates/ingress.yaml
+++ b/charts/prefect-orion/templates/ingress.yaml
@@ -20,7 +20,7 @@ metadata:
     {{- end }}
   {{- end }}
 spec:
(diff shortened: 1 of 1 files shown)


jamiezieziula (approved):
good catch!

PR #65, opened 2022-09-28
fix: Update servicePortName, doesn't fit on ingress

blopezpi:
Hi Prefect team, 

I'm working on deploying your helm chart into a kubernetes environment but I have a simple problem with the ingress portName `http` is not the same as `orion-svc-port`, so the ingress cannot be attached with the service itself. 

I generate this PR to fix that, but a simple option is to provide `orion-svc-port` instead of `http` in the ingress.yaml template.

If you need to open an issue, I will be able to open it.

Thank you.

### DIFF
--- a/charts/prefect-orion/templates/_helpers.tpl
+++ b/charts/prefect-orion/templates/_helpers.tpl
@@ -9,6 +9,10 @@ Create the name of the service account to use
 {{- end -}}
 {{- end -}}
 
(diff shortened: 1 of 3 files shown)


jamiezieziula:
Thanks @blopezpi - I opened this PR with the simple fix of replacing `http` with `orion-svc-port`.  Thanks for finding!

PR #71, opened 2022-10-06
feat: add support for passing additional permissions to the agent role

### LINKED ISSUE #70: Cannot create resource pods when using Dask
If we define a flow with a Dask task runner and `dask_kubernetes.KubeCluster` as its `cluster_class`, the agent cannot create pods in the cluster with the error:
```
"message":"pods is forbidden: User \"system:serviceaccount:myserviceaccount\" cannot create resource \"pods\" in API group \"\" in the namespace \"mynamespace\"","reason":"Forbidden","details":{"kind":"pods"},"code":403
```

I can obviously add my own `role` and `rolebinding` for the service account created in the `prefect-agent` helm chart but maybe you were planning to support this?

jamiezieziula:
Closes #70

### DIFF
--- a/charts/prefect-agent/values.yaml
+++ b/charts/prefect-agent/values.yaml
@@ -65,6 +65,8 @@ agent:
     # -- prefect API url
     apiUrl: http://127.0.0.1:4200/api
 
+  # -- enable this to grant additional permission to the agent role if you wish to use the Dask task runner
+  daskKubernetesCluster: false
 
   # -- number of agent replicas to deploy
   replicaCount: 1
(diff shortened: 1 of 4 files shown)


charts/prefect-agent/values.yaml
```diff
@@ -65,6 +65,8 @@ agent:
     # -- prefect API url
     apiUrl: http://127.0.0.1:4200/api
 
+  # -- enable this to grant additional permission to the agent role if you wish to use the Dask task runner
```
madkinsz, line 68: I'm hesitant because this is such a specific use case, but I guess the cost is low and people using Prefect on K8s are the ones likely to also use Dask on K8s 🤷‍♀️ 
```suggestion
  # -- enable this to grant additional permission to the agent role if you are using the Dask task runner to create dynamic Dask Kubernetes pods 
```

zanieb (approved):
Looks great!

PR #77, opened 2022-11-16
chore: replace MARVIN_GITHUB_PAT with local token

parkedwards:
https://docs.github.com/en/actions/security-guides/automatic-token-authentication

the workflow, which autogenerates a temp token to be used during the jobs, has these permissions scoped:

```
permissions:
  contents: write
  pull-requests: write
```


jamiezieziula (approved):

jawnsy (approved):
Awesome!

PR #79, opened 2022-11-25
Complete extraTls support

gregopenit:
Complete the support of the extraTls value which appears in the ingress.yaml. This is for example required to specify a custom secretName for the Tls certificate. For instance:
```
  tls: true
  extraTls:
    - hosts:
      - myhost.mydomain
      secretName: ingress-tls-myhost
```

Also clarify or correct some comments.

### DIFF
--- a/charts/prefect-orion/README.md
+++ b/charts/prefect-orion/README.md
@@ -37,6 +37,7 @@ Prefect orion application bundle
 | ingress.extraHosts | list | `[]` | an array with additional hostname(s) to be covered with the ingress record |
 | ingress.extraPaths | list | `[]` | an array with additional arbitrary paths that may need to be added to the ingress under the main host |
 | ingress.extraRules | list | `[]` | additional rules to be covered with this ingress record |
+| ingress.extraTls | list | `[]` | an array with additional tls configuration to be added to the ingress record |
 | ingress.host.hostname | string | `"prefect.local"` | default host for the ingress record |
 | ingress.host.path | string | `"/"` | default path for the ingress record |
 | ingress.host.pathType | string | `"ImplementationSpecific"` | ingress path type |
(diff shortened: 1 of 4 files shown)


gregopenit:
There are two secrets which will be created when `postgresql.auth.existingSecret` is not specified:
- {release-name}-postgresql which contains the keys `password` and `postgres-password`, created and managed by postgresql subchart
- {release-name}-postgresql-connection which contains the key `connection-string`, created and managed by prefect chart. 

Several issues exists with the management of these secrets:
- The description of the helper `orion.postgres-string-secret-name` mentions the first secret when actually defining  the name of the second secret.
- The values under `postgres` key are intended to the postgres subchart but are used by the main chart.
- If the`existingSecret` key is defined to allow the postgresql subchart to load the keys `password` and `postgres-password`, then the main chart will also try to load the key `connection-string` from this secret. But initially these keys are managed by two different charts in two different secrets.

I suggest then to dedicate a new value to define the secret containing the key `connection-string`.

jamiezieziula:
Thanks @gregopenit - i'll test this soon and merge if everything looks good!

jamiezieziula:
Hi @gregopenit, can you either provide documentation of the a successful deployment with your changes, or follow [these](https://github.com/PrefectHQ/prefect-helm#opening-a-pr) docs to allow Prefect Helm maintainers to checkout your forked branch?

gregopenit:
Helloe @jamiezieziula, my forked branch is public and I also allowed edits by the maintainers of the upstream branch. What is still missing ?
I can't provide a successful pipeline as I am using an internal CI pipeline solution but the example in description is an
anonymized version on my additional values in my deployment. The spec of the ingress resources including the extraTls section is something like:
```
spec:
  ingressClassName: nginx
  tls:
    - hosts:
        - myhost.mydomain
      secretName: ingress-tls-myhost
  rules:
    - host: myhost.mydomain
      http:
        paths:
          - path: /
            pathType: ImplementationSpecific
            backend:
              service:
                name: prefect-orion
                port:
                  name: orion-svc-port
```

jamiezieziula (approved):

jamiezieziula:
Thanks @gregopenit for your contribution - everything looks good on my end!

PR #90, opened 2022-12-09
chore: add default resource request/limits

### LINKED ISSUE #85: Add default resources
Users frequently ask for our recommendation on resource requests/limits for prefect agents. As of right now the numbers that I have are:

2048 CPU and 16000(aprox) memory

jawnsy:
Add resource requests and limits using figures from our internal data warehouse. Our agents use about 128 MiB, and a very small amount of CPU, so this should provide a reasonable starting point.

Closes: #85


parkedwards (approved):
yeah this seems a lot more reasonable to me

jamiezieziula (approved):

PR #94, opened 2022-12-10
fix: avoid installing postgresql unless postgresql.enabled

### LINKED ISSUE #93: Force useSubChart to false if postgres.enabled is false
**Summary**

If users disable PostgreSQL, the default behavior is to install PostgreSQL from the subchart anyways.

**Observed behavior**

Run `helm install orion prefect/prefect-orion --set=postgresql.enabled=false`, and you will have a PostgreSQL installation.

**Expected behavior**

When suppressing use of PostgreSQL, we should default to SQLite and thus have no need to install the PostgreSQL chart. Thus we should avoid installing PostgreSQL if `postgresql.enabled` is false.

**Debugging**

You can observe this behavior by running the following command: `helm template orion ./charts/prefect-orion --set=postgresql.enabled=false` from the repository root:

```
# Source: prefect-orion/charts/postgresql/templates/primary/statefulset.yaml
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: orion-postgresql
  namespace: "prefect-orion"
  labels:
    app.kubernetes.io/name: postgresql
    helm.sh/chart: postgresql-11.6.26
    app.kubernetes.io/instance: orion
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/component: primary
```

Users have to specify both `postgres.enabled=false` and `postgresql.useSubChart=false` to get an installation of Orion backed by SQLite only.

jawnsy:
By default, users that set postgresql.enabled to false still install PostgreSQL from the subchart, even though it is not used.

Update the subchart condition to avoid installing the subchart unless both postgresql.enabled and postgresql.useSubChart are true.

Closes: #93

### DIFF
--- a/charts/prefect-orion/Chart.yaml
+++ b/charts/prefect-orion/Chart.yaml
@@ -7,7 +7,7 @@ dependencies:
       - bitnami-common
     version: 2.2.1
   - name: postgresql
-    condition: postgresql.useSubChart
+    condition: postgresql.enabled,postgresql.useSubChart
     repository: https://charts.bitnami.com/bitnami
     version: 11.9.13
 description: Prefect orion application bundle


zanieb (approved):

charts/prefect-orion/Chart.yaml
```diff
@@ -7,7 +7,7 @@ dependencies:
       - bitnami-common
     version: 2.2.1
   - name: postgresql
-    condition: postgresql.useSubChart
+    condition: postgresql.enabled,postgresql.useSubChart
```
jawnsy, line 10: From the docs:

>Condition - The condition field holds one or more YAML paths (delimited by commas). If this path exists in the top parent's values and resolves to a boolean value, the chart will be enabled or disabled based on that boolean value. Only the first valid path found in the list is evaluated and if no paths exist then the condition has no effect.

https://helm.sh/docs/topics/charts/#tags-and-condition-fields-in-dependencies

I tested this by using `helm template orion`, before:

```
$ helm template orion . --set=postgresql.enabled=false | grep postgres
# Source: prefect-orion/charts/postgresql/templates/secrets.yaml
  name: orion-postgresql
    app.kubernetes.io/name: postgresql
    helm.sh/chart: postgresql-11.9.13
  postgres-password: "RVFhYTVjM1ZnSg==
```

After, the output is empty.

parkedwards (approved):

charts/prefect-orion/Chart.yaml
jamiezieziula: Do we need to still have the `useSubChart` value? or is its purpose duplicated?

jamiezieziula: ah i see - this makes sense to me!

jamiezieziula: @jawnsy I'm not sure this will work as expected. As you note - the condition only evaluates the `first` valid yaml path.  So if users enable `postgres` but do not want to use the `subchart`, this logic will not work for them

jawnsy: @jamiezieziula ah, good catch! you read the docs more closely than I did.

any thoughts what we should do to handle this? the condition field doesn't seem to support an "and" operator, as far as I can tell

jamiezieziula: I think we should move this logic elsewhere and leave `postgersql.useSubChart` as the condition on the chart itself. Let me take a look and get back to you!

jamiezieziula:
Going to close this for now - will reopen if it becomes more of a priority

PR #100, opened 2022-12-20
fix: support specifying multiple workqueues

### LINKED ISSUE #99: Specifying list of queues with agent's chart has unexpected behaviour
In the agent's helm chart, it's currently not possible to specify [several queues](https://github.com/PrefectHQ/prefect-helm/blob/f111427b1cd854e5b733f0d00c47a512a7f93006/charts/prefect-agent/values.yaml#L37-L39) as the list of queue names will just be concatenated, with a comma in between due to https://github.com/PrefectHQ/prefect-helm/blob/f111427b1cd854e5b733f0d00c47a512a7f93006/charts/prefect-agent/templates/deployment.yaml#L55
Is this known and will it be addressed soon? Thanks

jamiezieziula:
Resolves #99 
Screenshot 2022-12-20 at 12 03 07 PM

### DIFF
--- a/charts/prefect-agent/templates/deployment.yaml
+++ b/charts/prefect-agent/templates/deployment.yaml
@@ -56,8 +56,10 @@ spec:
             - prefect
             - agent
             - start
-            - -q
-            - {{ join " " .Values.agent.config.workQueues }}
+            {{- range .Values.agent.config.workQueues }}
+            - --work-queue
+            - {{ . | quote }}
+            {{- end }}
           workingDir: /home/prefect
           env:
             - name: HOME
--- a/charts/prefect-agent/templates/deployment.yaml
+++ b/charts/prefect-agent/templates/deployment.yaml
@@ -52,7 +52,12 @@ spec:
         - name: prefect-agent
           image: "{{ .Values.agent.image.repository }}:{{ .Values.agent.image.prefectTag }}"
           imagePullPolicy: {{ .Values.agent.image.pullPolicy }}
-          command: ["prefect", "agent", "start", "-q", '{{ join "," .Values.agent.config.workQueues }}']
+          command:
+            - prefect
+            - agent
+            - start
+            - -q
+            - {{ range .Values.agent.config.workQueues }} {{ . }}{{ end }}
           workingDir: /home/prefect
           env:
             - name: HOME
--- a/charts/prefect-agent/templates/deployment.yaml
+++ b/charts/prefect-agent/templates/deployment.yaml
@@ -57,7 +57,7 @@ spec:
             - agent
             - start
             - -q
-            - {{ range .Values.agent.config.workQueues }} {{ . }}{{ end }}
+            - {{ join " " .Values.agent.config.workQueues }}
           workingDir: /home/prefect
           env:
             - name: HOME


charts/prefect-agent/templates/deployment.yaml
```diff
@@ -52,7 +52,12 @@ spec:
         - name: prefect-agent
           image: "{{ .Values.agent.image.repository }}:{{ .Values.agent.image.prefectTag }}"
           imagePullPolicy: {{ .Values.agent.image.pullPolicy }}
-          command: ["prefect", "agent", "start", "-q", '{{ join "," .Values.agent.config.workQueues }}']
+          command:
+            - prefect
+            - agent
+            - start
+            - -q
+            - {{ range .Values.agent.config.workQueues }} {{ . }}{{ end }}
```
jawnsy, line 60: is this equivalent to doing: `{{ join " " .Values.agent.config.workQueues }}`?

also wondering if the fix is just to drop the single quotes we had before?

seems potentially useful for the agent to accept comma-delimited 🤔 

work queues support spaces in the name, so splitting on spaces doesn't seem like a reliable approach

This one probably needs some input from @madkinsz

jamiezieziula: It does not accept comma-delimited afaict
```
  ___ ___ ___ ___ ___ ___ _____     _   ___ ___ _  _ _____
 | _ \ _ \ __| __| __/ __|_   _|   /_\ / __| __| \| |_   _|
 |  _/   / _|| _|| _| (__  | |    / _ \ (_ | _|| .` | | |
 |_| |_|_\___|_| |___\___| |_|   /_/ \_\___|___|_|\_| |_|


Agent started! Looking for work from queue(s): default,test...
12:07:58.985 | INFO    | prefect.agent - Created work queue 'default,test'.
```

jamiezieziula: >work queues support spaces in the name, so splitting on spaces doesn't seem like a reliable approach

I'm not sure what you mean- neither the `join` nor the `range` function are splitting on spaces.

madkinsz: I don't see a strong justification for adding comma delimiters, best practice is

```
prefect agent start -q 'a' -q 'b'
```

I was surprised that it works with `-q a b` — that's cool. For spaces...

```
prefect agent start -q "with a space" "another"
```

jawnsy: Does something like this work?

```yaml
            {{ range .Values.agent.config.workQueues }}
            - -q
            - {{ . }}
            {{ end }}
```

I'm not sure if the agent is designed to accept multiple `-q` parameters, but that seems to be the way we can combine quoting (so we can handle spaces in queue names) and multiple queues?

Our code uses typer to handle it: https://github.com/PrefectHQ/prefect/blob/faae0f02b59a8dc83b4467e71b7f1efdae0d9ef7/src/prefect/cli/agent.py#L48-L53 and according to the example, we can pass the parameter multiple times: https://typer.tiangolo.com/tutorial/multiple-values/multiple-options/

I'd also suggest using the long-form name (--work-queue) because it makes it more obvious what's happening

jawnsy: > I'm not sure what you mean- neither the join nor the range function are splitting on spaces.

Yes, but the agent has to split on spaces to differentiate them, unless quoted as in Michael's example

But the "best practice" seems better anyway -- range with `-q` or `--work-queue` specified multiple times

jawnsy (approved):

jimid27 (approved):

PR #110, opened 2023-01-20
add namespace to role binding

acookin:
The rolebinding resource in the chart was missing the namespace in it's metadata.


jamiezieziula (approved):
nice find - thanks!

PR #111, opened 2023-01-28
clarify PREFECT_API_URL and PREFECT_ORION_UI_API_URL

### LINKED ISSUE #109: Update documentation for `PREFECT_API_URL`
Related to https://github.com/PrefectHQ/prefect-helm/issues/108 and other community messages, the documentation is not clear on what users should set the value of `PREFECT_API_URL` to.

AmyAiyyoi:
resolves [Update documentation for PREFECT_API_URL #109](https://github.com/PrefectHQ/prefect-helm/issues/109)

### DIFF
--- a/charts/prefect-agent/values.yaml
+++ b/charts/prefect-agent/values.yaml
@@ -65,7 +65,7 @@ agent:
     cloudUrl: https://api.prefect.cloud/api
 
   orionApiConfig:
-    # -- prefect API url; used by orion UI; should be set to publicly accessible API url
+    # -- prefect API url (PREFECT_API_URL); should be in-cluster url if the agent is deployed in the same cluster as the API
     apiUrl: http://127.0.0.1:4200/api
 
   # -- number of agent replicas to deploy
--- a/charts/prefect-agent/values.yaml
+++ b/charts/prefect-agent/values.yaml
@@ -65,7 +65,7 @@ agent:
     cloudUrl: https://api.prefect.cloud/api
 
   orionApiConfig:
-    # -- prefect API url (PREFECT_API_URL); should be in-cluster url if the agent is deployed in the same cluster as the API
+    # -- prefect API url (PREFECT_API_URL); should be in-cluster URL if the agent is deployed in the same cluster as the API
     apiUrl: http://127.0.0.1:4200/api
 
   # -- number of agent replicas to deploy
--- a/charts/prefect-agent/values.yaml
+++ b/charts/prefect-agent/values.yaml
@@ -65,7 +65,7 @@ agent:
     cloudUrl: https://api.prefect.cloud/api
 
   orionApiConfig:
-    # -- prefect API url
+    # -- prefect API url; used by orion UI; should be set to publicly accessible API url
     apiUrl: http://127.0.0.1:4200/api
 
   # -- number of agent replicas to deploy
--- a/charts/prefect-orion/values.yaml
+++ b/charts/prefect-orion/values.yaml
@@ -38,8 +38,8 @@ orion:
   # -- array with environment variables to add to orion nodes
   env: []
   ## env:
-  ##   - name: PREFECT_API_URL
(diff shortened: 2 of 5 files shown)


charts/prefect-agent/values.yaml
```diff
@@ -65,7 +65,7 @@ agent:
     cloudUrl: https://api.prefect.cloud/api
 
   orionApiConfig:
-    # -- prefect API url
+    # -- prefect API url; used by orion UI; should be set to publicly accessible API url
```
madkinsz, line 68: Since this is in the agent chart, it should just be the API URL not the UI API URL right?

charts/prefect-orion/values.yaml
```diff
@@ -35,6 +35,7 @@ orion:
   # -- array with environment variables to add to orion nodes
   env: []
   ## env:
+  ##   -- API url used by services such as the agent; can be in-cluster API url
```
madkinsz, line 38: Since this is in the Orion chart (not the agent chart) does this make sense? Is the UI API URL referenced anywhere in this chart? This is where it's relevant.

charts/prefect-orion/values.yaml
```diff
@@ -32,10 +32,12 @@ orion:
     # -- enable orion image debug mode
     debug: false
 
+  # -- PREFECT_ORION_UI_API_URL; should be publically accesible API url; default to PREFECT_API_URL for running API locally
```
madkinsz, line 35: I'm not sure "default to PREFECT_API_URL for running API locally" applies here since the API is running in a cluster by definition here?

charts/prefect-orion/values.yaml
```diff
@@ -32,10 +32,12 @@ orion:
...
+  publicApiUrl: ""
+
   # -- array with environment variables to add to orion nodes
   env: []
   ## env:
-  ##   -- API url used by services such as the agent; can be in-cluster API url
   ##   - name: PREFECT_API_URL
   ##     value: http://prefect-orion:4200/api
```
madkinsz, line 42: What nodes would these be? A better example setting may be changing the logging level.

charts/prefect-orion/values.yaml
madkinsz: May also want to note that if a value is not provided the UI will not work.

madkinsz: Yeah but they won't have set `PREFECT_API_URL` and they shouldn't here so it just adds confusion

madkinsz: `PREFECT_API_URL` is null by default so that it defaults to that value isn't really helpful for them here. They shouldn't use `PREFECT_API_URL` API in this context so I think it just adds confusion.

charts/prefect-agent/values.yaml
```diff
@@ -65,7 +65,7 @@ agent:
     cloudUrl: https://api.prefect.cloud/api
 
   orionApiConfig:
-    # -- prefect API url
+    # -- prefect API url (PREFECT_API_URL); should be in-cluster url if the agent is deployed in the same cluster as the API
```
tpdorsey, line 68:
```suggestion
    # -- prefect API url (PREFECT_API_URL); should be in-cluster URL if the agent is deployed in the same cluster as the API
```

charts/prefect-orion/values.yaml
```diff
@@ -32,11 +32,14 @@ orion:
     # -- enable orion image debug mode
     debug: false
 
+  # -- sets PREFECT_ORION_UI_API_URL; should be publically accesible API url; UI will not work unless set
```
tpdorsey, line 35:
```suggestion
  # -- sets PREFECT_ORION_UI_API_URL; should be publicly accessible API URL; UI will not work unless set
```

tpdorsey (dismissed):
Looks good. Minor suggested edits

parkedwards (dismissed):

jamiezieziula (approved):

PR #119, opened 2023-02-16
correct condition for tls spec in ingress resource

gregopenit:
Rollback of a last minute change in the PR #79. See my last comments.

@jamiezieziula has changed the condition in ingress.yaml which breaks the logic. With the current values:

```
ingress:
  enabled: true
  className: "nginx"
  host:
    hostname: myhost.mydomain
  tls: true
  extraTls:
    - hosts:
      - myhost.mydomain
      secretName: ingress-tls-myhost
```

Both conditions line 55 and line 60 would match, resulting in:

```
- hosts:
    - "myhost.mydomain"
  secretName: myhost.mydomain-tls
- hosts:
    - myhost.mydomain
  secretName: ingress-tls-myhost
```

Which does not make sense. Onemay want to define `.Values.ingress.extraTls` in order to set a specific `secretName` .


jamiezieziula (approved):
Ah I see - good call @gregopenit

PR #121, opened 2023-02-27
Adds option to specify workPool

EmilRex:
Adds option to set a single `workPool` in chart.

Still need to test the change to verify that existing behavior isn't changed and new behavior is.

### DIFF
--- a/charts/prefect-agent/README.md
+++ b/charts/prefect-agent/README.md
@@ -39,8 +39,8 @@ Prefect Agent application bundle
 | agent.clusterUid | string | `""` | unique cluster identifier, if none is provided this value will be infered at time of helm install |
 | agent.config.prefetchSeconds | int | `10` | when querying for runs, how many seconds in the future can they be scheduled |
 | agent.config.queryInterval | int | `5` | how often the agent will query for runs |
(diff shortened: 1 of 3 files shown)


jamiezieziula (dismissed):
lgtm

parkedwards (dismissed):

jimid27 (dismissed):
nice looks good

jicuss:
this would be super helpful to me

Bada-S:
Hey y'all, I tested out the chart locally. I pulled down Emil's chart and deployed the agent on minikube. I ended up hardcoding the work pool for the sake of testing, but it was able to deploy the agent successfully. @EmilRex If you can sync sometime I can show you how I tested the chart

EmilRex:
Alright... I made (and tested) some changes to the default behavior. Now if neither `workPool` nor `workQueues` are specified, we just listen to the default work queue. This ensures that the default behavior of the chart remains unchanged. Then, if either (or both) `workPool` or `workQueues` is specified, we fill them in as appropriate.

@jimid27 Does that make sense to you?

jimid27 (approved):
This lgtm!

PR #130, opened 2023-03-21
Look for agent.clusterUID instead of agent.clusterUUID in the deployment spec

jexxers:
I think this is a typo.

Ref: https://github.com/PrefectHQ/prefect-helm/blob/98f7de9fdaeb09418782cf5864f564ffe9ad788e/charts/prefect-agent/templates/deployment.yaml#L83

### DIFF
--- a/charts/prefect-agent/templates/deployment.yaml
+++ b/charts/prefect-agent/templates/deployment.yaml
@@ -80,7 +80,7 @@ spec:
             - name: PREFECT_API_URL
               value: {{ template "agent.apiUrl" . }}
             - name: PREFECT_KUBERNETES_CLUSTER_UID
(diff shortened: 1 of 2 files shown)


jexxers:
FWIW - the agent is looking for `clusterUID`.

Ref:
https://github.com/PrefectHQ/prefect/blob/42ab288f86f20cbfd804c9073d4df5a4a6a88299/src/prefect/infrastructure/kubernetes.py#L416

jexxers:
https://github.com/PrefectHQ/prefect-helm/blob/abe3dd1ce7979102b0f7209d77d04ad57f5c961a/charts/prefect-agent/templates/_helpers.tpl#L28-L35

Nevermind, I think all is good.

PR #133, opened 2023-03-27
feat: add experimental worker chart

jawnsy:
* Update Helm installed in build from 3.9.2 to 3.11.2
* Update agent/worker installation arguments from legacy "config" to new "cloudApiConfig" value
* Update agent values docstring to match generated helm-docs


jamiezieziula (approved):
nice!

PR #140, opened 2023-04-03
feat: allow configuration of http2 connections

jawnsy:
Add a value to configure the use of HTTP/2 if supported by the server, and disable it by default, as there are outstanding issues with httpx's support of HTTP/2.

### DIFF
--- a/charts/prefect-agent/values.yaml
+++ b/charts/prefect-agent/values.yaml
@@ -48,7 +48,7 @@ agent:
     # -- when querying for runs, how many seconds in the future can they be scheduled
     prefetchSeconds: 10
     # -- connect using HTTP/2 if the server supports it (experimental)
-    http2: false
+    http2: true
 
   ## connection settings
   # -- one of 'cloud' or 'server'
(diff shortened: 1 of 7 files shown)


EmilRex:
@jawnsy I love this! HTTP/2 issues are an incredibly common source of errors and confusion!

jamiezieziula (dismissed):
I'm fine with this - but I think it would be worth noting in the release notes that this reverese previous default behavior OR updating the default value to align with what was being applied by default previously

charts/prefect-agent/values.yaml
```diff
@@ -47,6 +47,8 @@ agent:
     queryInterval: 5
     # -- when querying for runs, how many seconds in the future can they be scheduled
     prefetchSeconds: 10
+    # -- connect using HTTP/2 if the server supports it (experimental)
+    http2: true
```
jawnsy, line 51: I went with the existing default of enabling HTTP/2 for now

jamiezieziula (approved):

PR #144, opened 2023-04-07
feat: default to Chart AppVersion as tag

jawnsy:
* Add standard app.kubernetes.io/version label
* Use the chart's appVersion as the image tag if prefectTag is not set, simplifying the chart packaging process


jawnsy:
This was a bad idea

PR #161, opened 2023-05-02
set http2 to default false

akdienes:
https://github.com/PrefectHQ/prefect/issues/7442
https://github.com/encode/httpx/discussions/2112

considering that the above two issues mean that using `http2: true` leads consistently to irrecoverable worker failures, I think the default should be false

### DIFF
--- a/charts/prefect-worker/values.yaml
+++ b/charts/prefect-worker/values.yaml
@@ -45,7 +45,7 @@ worker:
     # -- when querying for runs, how many seconds in the future can they be scheduled
     prefetchSeconds: 10
     # -- connect using HTTP/2 if the server supports it (experimental)
-    http2: true
+    http2: false
 
   ## connection settings
   # -- one of 'cloud' or 'server'


akdienes:
Hi @jawnsy , thanks for the quick response!

I was about 2 weeks behind latest, which looks like right around when that PR was merged. I'll update that and let you know if the worker failures persist or not

akdienes:
just wanted to update that I attempted to update the chart, but it causes the worker to fail in kube with
```prefect.exceptions.PrefectHTTPStatusError: Client error '404 Not Found' for url 'https://prefect.manifoldtrading.net/api/deployments/89ee0166-b731-4967-8d0b-19d7d3a6bb2e'```

Indeed this deployment does not exist. I'm not sure where it's getting that from or why it's trying to query that. moreover, the prefect UI shows the worker as healthy despite crashing repeatedly in kube

anyway, I've reverted back to the chart from mar 10 and the above error as gone away. just setting `http2: false` locally for the time being

jawnsy:
This was an issue with an upstream dependency, which should have been [resolved in encode/httpcore 0.17.1](https://github.com/encode/httpcore/pull/679). I'm closing this out as the default behavior (to use HTTP/2) should now work reliably. Please let us know if you see otherwise after upgrading to the latest version.

PR #167, opened 2023-05-25
chore: add liveness/readiness configs to prefect-server

### LINKED ISSUE #132: Prefect server should have liveness and readyness probe
As a DevOps team relying on Kubernetes to give us the health of the system, we need to have Prefect to provide us with the status of the pod using the [liveness and readiness probe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/)

At the moment, the Prefect server is missing endpoints like 
- ping => used for liveness
- health => used for readiness

I'm also raising an issue on the Prefect repository to get these endpoints added.

parkedwards:
when `readinessProbe.enabled` is `true`, the tests will fail as expected because the endpoint will go out in the upcoming `prefect` release

resolves https://github.com/PrefectHQ/prefect-helm/issues/132


jamiezieziula (dismissed):

charts/prefect-worker/README.md
```diff
@@ -6,10 +6,6 @@ Prefect Worker application bundle
 
 **Homepage:** <https://github.com/PrefectHQ>
 
-## Getting Started Guide
-
-[Using Prefect Helm Chart for Prefect Workers](https://docs.prefect.io/latest/guides/deployment/helm-worker/)
```
parkedwards, line 11: @jamiezieziula any idea why the helm-docs pre-commit tries to scrub this?

charts/prefect-worker/README.md.gotmpl
```diff
@@ -0,0 +1,21 @@
...
+
+{{ template "chart.badgesSection" . }}
+
+{{ template "chart.description" . }}
+
+{{ template "chart.homepageLine" . }}
+
+## Getting Started Guide
+
+[Using Prefect Helm Chart for Prefect Workers](https://docs.prefect.io/latest/guides/deployment/helm-worker/)
```
parkedwards, line 11: we can augment the base template that `helm-docs` uses by adding this `.gotmpl` template, and the `{{ template }}` blocks ive added here are (1) from the library and (2) should match our existing autogenerated blocks, in case we ever want to augment the existing readme format for the other charts

jimid27 (approved):

PR #179, opened 2023-06-09
Add configuration for priority classes

### LINKED ISSUE #178: Ability to configure priority classes
We are running the server and agent in the same cluster. If there are too many jobs running and some pods have to be evicted to make room, we want the server pods to keep running. I found out that a way to do this is by using priority classes (https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#priorityclass). It would be nice to be able to configure this in values.yaml.

pravindahal:
Closes: #178


charts/prefect-agent/templates/deployment.yaml
```diff
@@ -48,6 +48,9 @@ spec:
       {{- if .Values.agent.podSecurityContext }}
       securityContext: {{- .Values.agent.podSecurityContext | toYaml | nindent 8 }}
       {{- end }}
+      {{- if .Values.server.priorityClassName }}
```
jawnsy, line 51: This looks like a typo

```suggestion
      {{- if .Values.agent.priorityClassName }}
```

jawnsy (commented):
Hey there, thanks so much for your contribution! This change looks good to me, however, there seem to be a few typos.

Can you please also re-generate our documentation using `helm-docs`?

jawnsy (approved):
LGTM, thanks for your contribution!

PR #196, opened 2023-07-05
fix/updatecli workflow secret

### LINKED ISSUE #173: Fix update workflow
Our updatecli workflow is failing due to a permissions issue. I think I've seen this before, and the problem is that we're trying to update a workflow from a running workflow.

This is the error:

```
 ! [remote rejected] helm-version-2023-06-01 -> helm-version-2023-06-01 (refusing to allow a GitHub App to create or update workflow `.github/workflows/agent-lint-and-test.yaml` without `workflows` permission)
```

We saw this previously when trying to update the gcloud sdk in our build: https://github.com/PrefectHQ/nebula/pull/4403 - this seems to have been fixed in https://github.com/PrefectHQ/nebula/pull/4401

Example of failing run: https://github.com/PrefectHQ/prefect-helm/actions/runs/5149664089

jimid27:
Resolves: https://github.com/PrefectHQ/prefect-helm/issues/173

### DIFF
--- a/.github/workflows/update-helm-minor-versions.yaml
+++ b/.github/workflows/update-helm-minor-versions.yaml
@@ -23,6 +23,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+          token: ${{ secrets.PREFECT_HELM_CONTENTS_WORKFLOWS_RW }}
 
       - name: configure git
         run: |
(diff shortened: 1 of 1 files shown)


jamiezieziula (approved):

.github/workflows/update-helm-minor-versions.yaml
```diff
@@ -23,6 +23,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+          token: ${{ secrets.PREFECT_HELM_CONTENTS_WORKFLOWS_RW }}
```
parkedwards, line 28: im not sure these additions are necessary on checkout

PR #207, opened 2023-07-19
fix: use appVersion from chart for label

### LINKED ISSUE #203: Add ability to pass in Image digest instead of tag
It would be helpful to pull a specific prefect docker image by digest rather than pulling the latest push to a tag. For example if I want to pull `prefecthq/prefect@sha1234567.....`, there should be a value in values.yaml `digest: sha1234567.....`. The current configuration only allows for the tag to be specified

jawnsy:
Use the chart's appVersion field rather than image tag for the prefect-version label, tags can exceed 63 characters when using image digests.

Resolves: #203

### DIFF
--- a/charts/prefect-agent/templates/deployment.yaml
+++ b/charts/prefect-agent/templates/deployment.yaml
@@ -5,7 +5,7 @@ metadata:
   namespace: {{ include "common.names.namespace" . | quote }}
   labels: {{- include "common.labels.standard" . | nindent 4 }}
     app.kubernetes.io/component: agent
-    prefect-version: {{ .Values.agent.image.prefectTag }}
+    prefect-version: {{ .Chart.AppVersion }}
     {{- if .Values.commonLabels }}
     {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" $ ) | nindent 4 }}
     {{- end }}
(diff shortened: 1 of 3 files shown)


charts/prefect-agent/templates/deployment.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
   namespace: {{ include "common.names.namespace" . | quote }}
   labels: {{- include "common.labels.standard" . | nindent 4 }}
     app.kubernetes.io/component: agent
-    prefect-version: {{ .Values.agent.image.prefectTag }}
+    prefect-version: {{ .Chart.AppVersion }}
```
jawnsy, line 8: Current behavior:

```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: prefect-worker
  namespace: "jawnsy-test"
  labels:
    app.kubernetes.io/name: prefect-worker
    helm.sh/chart: prefect-worker-0.0.0
    app.kubernetes.io/instance: worker
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/component: worker
    prefect-version: 2-python3.11-kubernetes
```

New behavior:

```yaml
apiVersion: v1
kind: Service
metadata:
  name: prefect-server
  namespace: "jawnsy-test"
  labels:
    app.kubernetes.io/name: prefect-server
    helm.sh/chart: prefect-server-0.0.0
    app.kubernetes.io/instance: server
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/component: server
    prefect-version: 2-latest
```

TODO: we should probably use [`app.kubernetes.io/version`](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/) for this instead

jamiezieziula (approved):

PR #209, opened 2023-07-19
chore: update dependencies

jawnsy:
Run updatecli manually, as there is currently a failure related to credentials that is preventing updatecli from running.

### DIFF
--- a/.github/workflows/agent-lint-and-test.yaml
+++ b/.github/workflows/agent-lint-and-test.yaml
@@ -1,12 +1,10 @@
 name: Lint and Test Prefect Agent Chart
-
 on:
   pull_request:
     paths:
       - .github/workflows/agent-lint-and-test.yaml
       - .github/linters/agent-ct.yaml
       - 'charts/prefect-agent/**'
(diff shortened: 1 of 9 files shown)


jimid27 (approved):

jamiezieziula (approved):
any reason for removing the new lines?

.github/workflows/agent-lint-and-test.yaml
```diff
@@ -1,12 +1,10 @@
 name: Lint and Test Prefect Agent Chart
-
 on:
   pull_request:
     paths:
       - .github/workflows/agent-lint-and-test.yaml
       - .github/linters/agent-ct.yaml
       - 'charts/prefect-agent/**'
-
```
jawnsy, line 9: @jamiezieziula updatecli seemed to remove these, and I have no idea how to control that behavior

jawnsy: Merging for now, if we figure out some way to fix updatecli, we can bring these back

PR #212, opened 2023-07-26
Add `workQueues` option to `worker` chart

### LINKED ISSUE #211: Add `workQueues` to prefect-worker Helm values
Looking to be able to configure which work queues the worker pulls from via Helm values, like we are currently able to do with prefect-agent. E.g.,

```yaml
spec:
  values:
    worker:
      config:
        workQueues:
          - my-work-queue
```

Thanks!

jamiezieziula:
Resolves https://github.com/PrefectHQ/prefect-helm/issues/211

### DIFF
--- a/charts/prefect-worker/templates/deployment.yaml
+++ b/charts/prefect-worker/templates/deployment.yaml
@@ -68,11 +68,9 @@ spec:
             - kubernetes
             - --pool
             - {{ .Values.worker.config.workPool | quote }}
-            {{- if .Values.worker.config.workQueues }}
             {{- range .Values.worker.config.workQueues }}
             - --work-queue
             - {{ . }}
-            {{- end }}
             {{- end }}
             {{- if .Values.worker.config.limit }}
             - --limit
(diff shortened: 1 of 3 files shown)


charts/prefect-worker/values.yaml
```diff
@@ -37,8 +37,10 @@ worker:
 
   ## general configuration of the worker
   config:
-    # -- name of prefect work pool the worker will poll
+    # -- the work pool the started worker should poll.
```
parkedwards, line 40:
```suggestion
    # -- the work pool that your started worker will poll.
```

parkedwards (dismissed):
tight

charts/prefect-worker/templates/deployment.yaml
```diff
@@ -68,6 +68,12 @@ spec:
             - kubernetes
             - --pool
             - {{ .Values.worker.config.workPool | quote }}
+            {{- if .Values.worker.config.workQueues }}
```
jawnsy, line 71: I don't think you need the `if` because a range over an empty list won't run (like a for-each over an empty array)

jawnsy (dismissed):

charts/prefect-worker/templates/deployment.yaml
jamiezieziula: nice !

jawnsy (approved):

PR #214, opened 2023-07-30
Enable exposing prefect-server with Loadbalancer

g-clef:
Also enable setting prefect's `APIHost` value and k8s service `targetPort` value. 

Example values.yaml to install with this (using a MetalLB LoadBalancer, and an existing, in-cluster Postgres db):
```
fullnameOverride: "prefect-server"
namespaceOverride: "prefect-server"

postgresql:
  enabled: true
  useSubChart: false
  auth:
    existingSecret: prefect-postgres-auth
  externalHostname: postgresql

service:
  type: LoadBalancer
  loadBalancerIP: 192.168.1.11
  annotations:
    metallb.universe.tf/address-pool: exposed-pool

server:
  publicApiUrl: "http://prefect-server.example.com:4200/api"
  apiHost: "prefect-server.example.com"
  image:
    prefectTag: 2.8-python3.11
```

### DIFF
--- a/charts/prefect-server/templates/deployment.yaml
+++ b/charts/prefect-server/templates/deployment.yaml
@@ -82,6 +82,10 @@ spec:
             - name: PREFECT_UI_API_URL
               value: {{ .Values.server.publicApiUrl | quote }}
             {{- end }}
+            {{- if .Values.server.apiHost }}
+            - name: PREFECT_SERVER_API_HOST
+              value: {{ .Values.server.apiHost | quote }}
+            {{- end }}
             {{- if .Values.postgresql.enabled }}
             - name: PREFECT_API_DATABASE_CONNECTION_URL
               valueFrom:
--- a/charts/prefect-server/templates/service.yaml
+++ b/charts/prefect-server/templates/service.yaml
@@ -26,6 +26,9 @@ spec:
   {{- if eq .Values.service.type "NodePort" }}
   externalTrafficPolicy: {{ .Values.service.externalTrafficPolicy | quote }}
   {{- end }}
+  {{- if eq .Values.service.type "LoadBalancer" }}
+  loadBalancerIP: {{ .Values.service.loadBalancerIP | quote }}
+  {{- end }}
   ports:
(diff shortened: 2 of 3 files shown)


charts/prefect-server/templates/deployment.yaml
```diff
@@ -82,6 +82,10 @@ spec:
             - name: PREFECT_UI_API_URL
               value: {{ .Values.server.publicApiUrl | quote }}
             {{- end }}
+            {{- if .Values.server.apiHost }}
+            - name: PREFECT_SERVER_API_HOST
```
jawnsy, line 86: Why is this setting necessary, can you explain?

The public API URL (the URL that the frontend uses to communicate with the backend) should be sufficient for most use cases.

We listen to all interfaces (albeit, currently, all IPv4 interfaces only) and that's set here: https://github.com/PrefectHQ/prefect-helm/blob/b694a97f614a52fdfa58cc8eca6faa59ebc22e77/charts/prefect-server/templates/deployment.yaml#L64-L68

You should be able to use a LoadBalancer service to route traffic to the OSS Prefect server without setting anything else

g-clef: It was a belt-and-suspenders thing. Based on https://github.com/PrefectHQ/prefect/blob/main/RELEASE-NOTES.md#20b5, since the API_HOST can be used to build the URL for API access, I wanted to make sure all possible paths for API URL building could be set to the proper host.

jamiezieziula:
@g-clef I've made some changes to the helm chart that might address the issues you originally faced. Can you confirm if they are resolved with the latest release (2023.10.17)?

g-clef:
Thanks. I'd hoped to be able to specify exposing the UI via a `LoadBalancer` like MetalLB (which was the main point of this PR). The other changes were mostly just to support pulling that off. If LoadBalancers aren't a thing you all want to support, that's fine.

jamiezieziula:
@g-clef ah I see! If you want to rebase your PR from main I'll give it review!

PR #227, opened 2023-09-06
enhancement: Add required functions to worker template

### LINKED ISSUE #223: Require specific values be provided in the prefect-worker helm chart
For workers being deployed in mode `cloud`, at a minimum the chart needs to receive the following values:
- work pool 
- account id 
- workspace id

We should require these at the helm level

jamiezieziula:
With the use of `{{ required }}`, we know enforce at the helm level the 3 necessary values needed in order for a worker to be deployed successfully.

This also fixes the printed UI url when using the `server` configuration which previously just displayed the `api url` endpoint. 

Resolves https://github.com/PrefectHQ/prefect-helm/issues/223

### DIFF
--- a/charts/prefect-worker/templates/_helpers.tpl
+++ b/charts/prefect-worker/templates/_helpers.tpl
@@ -14,7 +14,7 @@ Require Prefect Cloud Account ID
 */}}
 {{- define "cloud.requiredConfig.accountId" -}}
 {{- if eq .Values.worker.apiConfig "cloud" }}
-    {{- required "A Prefect Cloud Account ID is required (worker.cloudApiConfig.accountId)" .Values.worker.cloudApiConfig.accountId | printf "%s"  -}}
+    {{- required "A Prefect Cloud Account ID is required (worker.cloudApiConfig.accountId)" .Values.worker.cloudApiConfig.accountId -}}
 {{- end -}}
 {{- end -}}
 
@@ -23,7 +23,7 @@ Require Prefect Cloud Workspace ID
 */}}
 {{- define "cloud.requiredConfig.workspaceId" -}}
 {{- if eq .Values.worker.apiConfig "cloud" }}
-    {{- required "A Prefect Cloud Workspace ID is required (worker.cloudApiConfig.workspaceId)" .Values.worker.cloudApiConfig.workspaceId | printf "%s"  -}}
+    {{- required "A Prefect Cloud Workspace ID is required (worker.cloudApiConfig.workspaceId)" .Values.worker.cloudApiConfig.workspaceId -}}
 {{- end -}}
 {{- end -}}
 
--- a/charts/prefect-worker/templates/_helpers.tpl
+++ b/charts/prefect-worker/templates/_helpers.tpl
@@ -34,8 +34,10 @@ Require Prefect Cloud Workspace ID
 {{- define "worker.apiUrl" -}}
 {{- if eq .Values.worker.apiConfig "cloud" }}
     {{- printf "%s/accounts/%s/workspaces/%s" .Values.worker.cloudApiConfig.cloudUrl (include "cloud.requiredConfig.accountId" .) (include "cloud.requiredConfig.workspaceId" .) | quote }}
+{{- else if and (eq .Values.worker.apiConfig "server") .Values.worker.serverConfig.apiUrl.fullyQualifiedEndpoint }}
+    {{- printf .Values.worker.serverConfig.apiUrl.fullyQualifiedEndpoint | quote }}
 {{- else }}
-    {{- .Values.worker.serverApiConfig.apiUrl | quote }}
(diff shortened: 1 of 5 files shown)


charts/prefect-worker/templates/deployment.yaml
```diff
@@ -67,7 +67,7 @@ spec:
             - --type
             - {{ .Values.worker.config.type | quote }}
             - --pool
-            - {{ .Values.worker.config.workPool | quote }}
+            -  {{ required "A Work Pool Name is required (worker.config.workPool)" .Values.worker.config.workPool | quote }}
```
parkedwards, line 70: i can never tell - do we need a leading space hyphen thing here?

`{{- required`

parkedwards (approved):
damn, nice!

charts/prefect-worker/templates/deployment.yaml
jamiezieziula: it through some validation error when i had that in. good q tho!

charts/prefect-worker/templates/_helpers.tpl
```diff
@@ -9,15 +9,35 @@ Create the name of the service account to use
...
+    {{- required "A Prefect Cloud Account ID is required (worker.cloudApiConfig.accountId)" .Values.worker.cloudApiConfig.accountId | printf "%s"  -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Require Prefect Cloud Workspace ID
+*/}}
+{{- define "cloud.requiredConfig.workspaceId" -}}
+{{- if eq .Values.worker.apiConfig "cloud" }}
+    {{- required "A Prefect Cloud Workspace ID is required (worker.cloudApiConfig.workspaceId)" .Values.worker.cloudApiConfig.workspaceId | printf "%s"  -}}
```
parkedwards, line 26: can you remind me why we need to pipe the value into `printf "%s"`? isn't it already a string?

parkedwards: i get it for when we're constructing a bigger string w inputs, like the apiUrl below

parkedwards: oh do we need the quotes?

charts/prefect-worker/templates/_helpers.tpl
```diff
@@ -9,15 +9,35 @@ Create the name of the service account to use
...
 */}}
 {{- define "worker.apiUrl" -}}
 {{- if eq .Values.worker.apiConfig "cloud" }}
-{{- printf "%s/accounts/%s/workspaces/%s" .Values.worker.cloudApiConfig.cloudUrl .Values.worker.cloudApiConfig.accountId .Values.worker.cloudApiConfig.workspaceId | quote }}
+    {{- printf "%s/accounts/%s/workspaces/%s" .Values.worker.cloudApiConfig.cloudUrl (include "cloud.requiredConfig.accountId" .) (include "cloud.requiredConfig.workspaceId" .) | quote }}
+{{- else if and (eq .Values.worker.apiConfig "server") .Values.worker.serverConfig.apiUrl.fullyQualifiedEndpoint }}
+    {{- printf .Values.worker.serverConfig.apiUrl.fullyQualifiedEndpoint | quote }}
 {{- else }}
-{{- .Values.worker.serverApiConfig.apiUrl | quote }}
+    {{- printf "http://%s.%s.svc.cluster.local:%s/api" .Values.worker.serverConfig.apiUrl.prefectServer.serviceName .Values.worker.serverConfig.apiUrl.prefectServer.namespace .Values.worker.serverConfig.apiUrl.prefectServer.servicePort | quote}}
```
parkedwards, line 40: my opinion - i think the operator should just construct this k8s endpoint themselves, vs us handling this third case and having more value inputs overall.  so sharing the `fullyQualifiedEndpoint` input for both self-hosted or internal k8s

parkedwards (approved):
LGTM - i have mixed feelings about exposing extra k8s input values bc on the one hand, i like giving users a helping hand. on the other end, im worried this will confuse an operator on when to use that vs. a regular endpoint

charts/prefect-worker/templates/_helpers.tpl
jamiezieziula: ok this is a good question - let me test to see if we need the `printf` pipe

charts/prefect-worker/values.yaml
```diff
@@ -76,8 +76,12 @@ worker:
     cloudUrl: https://api.prefect.cloud/api
 
   serverApiConfig:
-    # -- prefect API url (PREFECT_API_URL); should be in-cluster URL if the worker is deployed in the same cluster as the API
-    apiUrl: http://127.0.0.1:4200/api
+    # If the prefect server is located external to this cluster, set a fully qualified domain name as the apiUrl
```
parkedwards, line 79: 👍

parkedwards (approved):

PR #228, opened 2023-09-07
Build out additional docs for the `prefect-worker` chart

jamiezieziula:


charts/prefect-worker/README.md
```diff
@@ -1,14 +1,121 @@
...
+    data:
+      key:  <base64-encoded-api-key>
+    ```
+
+    Replace `<base64-encoded-api-key>` with your Prefect Cloud API key encoded in base64. The helm chart looks for a secret of this name and schema, this can be overridden in the `values.yaml`.
+
+    You can use the following command to generate the base64-encoded value:
+
+    ```bash
+    echo -n "your-prefect-cloud-api-key" | base64
```
parkedwards, line 48: is there a missing step here? like getting the key into `api-key.yaml`

parkedwards: nvm, i missed the part right before where it says to set up the file

parkedwards (approved):

PR #255, opened 2023-10-05
Self hosted allow api key ref

jimid27:
- Need to allow api key ref when deploying to a self hosted instance of prefect

### DIFF
--- a/charts/prefect-worker/templates/deployment.yaml
+++ b/charts/prefect-worker/templates/deployment.yaml
@@ -93,12 +93,18 @@ spec:
               value: {{ template "worker.apiUrl" . }}
             - name: PREFECT_KUBERNETES_CLUSTER_UID
               value: {{ template "worker.clusterUUID" . }}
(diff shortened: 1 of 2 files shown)


charts/prefect-worker/templates/deployment.yaml
```diff
@@ -93,7 +93,7 @@ spec:
               value: {{ template "worker.apiUrl" . }}
             - name: PREFECT_KUBERNETES_CLUSTER_UID
               value: {{ template "worker.clusterUUID" . }}
-            {{- if eq .Values.worker.apiConfig "cloud" }}
+            {{- if or (eq .Values.worker.apiConfig "cloud") (eq .Values.worker.apiConfig "selfHosted") }}
             - name: PREFECT_API_KEY
               valueFrom:
                 secretKeyRef:
```
jamiezieziula, line 99: we'll need to handle this config also! you could create something similar to how we construct the api url in the helpers template for the secret name and key

jamiezieziula (changes requested):
use selfhosted secret name and key

jamiezieziula (approved):

PR #257, opened 2023-10-10
enhancement: add the ability to customize the target port of the server deployment

### LINKED ISSUE #256: Prefect server service targetPort
Normally with a service I'd specify a different port and leave targetPort as-is. I've tried changing the port to 80 but it causes weird errors with the prefect server. Currently I've patched the port of the service to 80 and left the targetPort to 4200 which works but is a manual process. 

I'd like to do this by specifying it in values.yaml but didn't find an option to specify different port/targetPort combinations.

jamiezieziula:
This PR allows end users to customize the exposed port of the `server` deployment. As a result, users have the ability to adjust the exposed service port without causing an issue with the `server` deployment

Resolves https://github.com/PrefectHQ/prefect-helm/issues/256


.github/workflows/deploy-latest-helm-version.yaml
```diff
@@ -10,6 +10,10 @@ jobs:
   update_helm_releases:
     runs-on: ubuntu-latest
     steps:
+      - name: Sleep for 3 minutes to allow time for the helm package to be available
```
parkedwards, line 13: ouch. so the package doesn't become available on the helm registry for a bit?

jamiezieziula: i think it just takes a minute or two

parkedwards (approved):

PR #259, opened 2023-10-16
Update environment variables passed to the prefect `server` chart

jamiezieziula:
This PR will implement a **breaking change**. The following will need to be updated in the end users environment when upgrading:

### Prefect API URL (server.prefectApiUrl)
This value was previously not exposed. It is set to default value that was previously set at`server.publicApiUrl`. If the URL of your server installation is not `http://localhost:4200`, you **will** need to override that value here.

### Prefect UI Configuration
Unless you have a different endpoint for your UI server, you can ignore this change. If your UI endpoint is unique from the API endpoint, you will need to set both `server.uiConfig.prefectUiApiUrl` and `server.uiConfig.prefectUiUrl`.

### Debug Level
If you previously had `server.image.debug` set to `true`, you will need to remove that setting and instead set `server.debug` to be `true`

See [here](https://docs.prefect.io/2.13.7/api-ref/prefect/settings/) for any questions around the possible server environment variables

### DIFF
--- a/charts/prefect-server/templates/deployment.yaml
+++ b/charts/prefect-server/templates/deployment.yaml
@@ -64,10 +64,6 @@ spec:
             - prefect
             - server
             - start
-            - --host
-            - 0.0.0.0
-            - --log-level
-            - WARNING
             - --port
             - {{ .Values.service.targetPort | quote }}
           workingDir: /home/prefect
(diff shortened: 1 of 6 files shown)


charts/prefect-server/values.yaml
```diff
@@ -39,6 +39,9 @@ server:
   # -- sets PREFECT_UI_API_URL; should be publicly accessible API URL
   publicApiUrl: http://localhost:4200/api
 
+  # -- sets PREFECT_UI_URL
+  publicUiUrl: "http://localhost:4200"
```
parkedwards, line 43: should we not set a default value, so that the behavior falls back to the underlying prefect library's behavior? genuinely not sure

charts/prefect-server/templates/deployment.yaml
```diff
@@ -76,11 +72,25 @@ spec:
...
             - name: PREFECT_DEBUG_MODE
-              value: {{ .Values.server.image.debug | quote }}
-            {{- if .Values.server.publicApiUrl }}
+              value: {{ .Values.server.debug | quote }}
+            - name: PREFECT_LOGGING_SERVER_LEVEL
+              value: {{ .Values.server.loggingLevel | quote }}
+            - name: PREFECT_SERVER_API_HOST
+              value: {{ .Values.server.prefectApiHost | quote }}
+            - name: PREFECT_SERVER_API_PORT
+              value: {{ .Values.service.targetPort  | quote }}
```
jamiezieziula, line 84: these were already being inferred or set elsewhere, so explicitly setting them here now

charts/prefect-server/templates/deployment.yaml
```diff
@@ -76,11 +72,25 @@ spec:
...
+            - name: PREFECT_UI_ENABLED
+              value: {{ .Values.server.uiConfig.enabled | quote }}
```
jamiezieziula, line 86: same thing here, this is the default behavior

charts/prefect-server/templates/deployment.yaml
```diff
@@ -76,11 +72,25 @@ spec:
           env:
             - name: HOME
               value: /home/prefect
+            - name: PREFECT_API_URL
+              value: {{ .Values.server.prefectApiUrl | quote }}
```
jamiezieziula, line 76: we were **not** previously setting this, which was incorrect behavior. All other URLs are inferred from this, so it **needs** to be set.

charts/prefect-server/templates/deployment.yaml
```diff
@@ -76,11 +72,25 @@ spec:
...
+            - name: PREFECT_UI_ENABLED
+              value: {{ .Values.server.uiConfig.enabled | quote }}
+            {{- if .Values.server.uiConfig.prefectUiApiUrl }}
             - name: PREFECT_UI_API_URL
-              value: {{ .Values.server.publicApiUrl | quote }}
+              value: {{ .Values.server.uiConfig.prefectUiApiUrl | quote }}
+            {{- end }}
+            {{- if .Values.server.uiConfig.prefectUiUrl }}
+            - name: PREFECT_UI_URL
+              value: {{ .Values.server.uiConfig.prefectUiUrl | quote }}
```
jamiezieziula, line 93: these values are **only** necessary if the endpoint for the UI is **different** than the URL set in the `prefectApiUrl` parameter

parkedwards (approved):

jakekaplan (approved):

PR #264, opened 2023-10-23
Provide guidance on running >1 worker in a single k8s namespace

jamiezieziula:
Resolves https://github.com/PrefectHQ/platform/issues/3708


parkedwards (approved):
so simple

PR #289, opened 2024-01-19
fix(server): mounted fs path for static-ui-dir

### LINKED ISSUE #288: Add PREFECT_UI_STATIC_DIRECTORY and emptyDir volume mount to fix Permission Error
Is needed to solve bug in prefect mentioned in https://github.com/PrefectHQ/prefect/issues/11640 and solved by prefect 2.14.16

Since image is run as non-root, the chart should provide read/write for the PREFECT_UI_STATIC_DIRECTORY dir. 

Set the env var PREFECT_UI_STATIC_DIRECTORY to the corresponding directory in the chart as well.

Am on mobile right now but will provide more information later if needed

parkedwards:
resolves https://github.com/PrefectHQ/prefect-helm/issues/288

image

```
$ kl prefect-server-5dd4b5755-7b5k7

 ___ ___ ___ ___ ___ ___ _____
| _ \ _ \ __| __| __/ __|_   _|
|  _/   / _|| _|| _| (__  | |
|_| |_|_\___|_| |___\___| |_|

Configure Prefect to communicate with the server with:

    prefect config set PREFECT_API_URL=http://0.0.0.0:4200/api

View the API reference documentation at http://0.0.0.0:4200/docs

Check out the dashboard at http://0.0.0.0:4200
```

### DIFF
--- a/charts/prefect-server/templates/deployment.yaml
+++ b/charts/prefect-server/templates/deployment.yaml
@@ -95,6 +95,8 @@ spec:
             - name: PREFECT_UI_URL
               value: {{ .Values.server.uiConfig.prefectUiUrl | quote }}
             {{- end }}
+            - name: PREFECT_UI_STATIC_DIRECTORY
+              value: {{ .Values.server.uiConfig.prefectUiStaticDirectory | quote }}
             {{- if .Values.postgresql.enabled }}
             - name: PREFECT_API_DATABASE_CONNECTION_URL
               valueFrom:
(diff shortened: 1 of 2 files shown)


charts/prefect-server/values.yaml
```diff
@@ -54,6 +54,8 @@ server:
     prefectUiApiUrl: ""
     # -- sets PREFECT_UI_URL
     prefectUiUrl: ""
+    # -- sets PREFECT_UI_STATIC_DIRECTORY
+    prefectUiStaticDirectory: "/ui_build"
```
parkedwards, line 58: this needs a default value, or else the server container will look for a non-existent default that is baked into the application code.  so the expectation is, when using the helm chart, we will be setting this by either user-provided value or this default value

jimid27 (approved):
nice

charts/prefect-server/templates/deployment.yaml
```diff
@@ -141,6 +143,9 @@ spec:
             - mountPath: /tmp
               name: scratch
               subPathExpr: tmp
+            - mountPath: {{ .Values.server.uiConfig.prefectUiStaticDirectory }}
+              name: scratch
```
parkedwards, line 147: nothing really special about this directory, so we can re-use the emptyDir that we initialize here

jamiezieziula (approved):

Issue #297, opened 2024-02-09
prefect-helm - support multiple OSS releases in a day


parkedwards:
we have to switch our release versioning schema to something that is similar to what [we have in nebula](https://github.com/PrefectHQ/nebula/blob/main/nebula-release#L15)

we'll also need to have it invokable by both OSS and in the helm repo itself (for ad-hoc or bugfix releases)

PR #298, opened 2024-02-09
chore: switch release workflow to manual, dispatch-based trigger

parkedwards:
Update 2/9/24 - we're also going to _**always**_ run the lint-and-test jobs, as they are required status checks in this repo

--
relates to https://github.com/PrefectHQ/prefect-helm/issues/297

currently, our chart release workflow looks like this:
1. a chart version is set according to the date `2024.2.8`
2. a release/tag in this repo triggers the helm build

a new chart release is triggered by one of the following:
1. [anytime our upstream OSS library has a new update](https://github.com/PrefectHQ/prefect/blob/main/.github/workflows/helm-chart-release.yaml)
2. a manual release/tag is created in this repo

this poses an issue whenever we need to cut multiple OSS releases on the same day, as this results in a release/tag conflict in this repo

in order to solve this, we will:
1. modify the chart versioning schema to include the numerical hour + minute + seconds via [pre-release append](https://semver.org/#spec-item-9) (at the time the release workflow is executed).  This allows us to create multiple releases on the same day, without fear of conflict, as the version schema will now be something like `2024.2.8-204348`
3. only run the release workflow on `workflow_dispatch`, which is reflects the current manual trigger for ad-hoc releases.  additionally, the upstream OSS repo will call `gh workflow run` on the `helm-release.yaml` workflow in this repo -- this ensures that both release triggers operate on the same logic

additionally, we'll avoid the pitfall revealed in [this issue](https://github.com/PrefectHQ/prefect-helm/pull/234) around semver-based ordering not honoring our versioning schema, as this pre-release append ascends numerically + honors the overall schema

https://jubianchi.github.io/semver-check/#/version/2024.2.8-204348

### DIFF
--- a/.github/workflows/helm-release.yaml
+++ b/.github/workflows/helm-release.yaml
@@ -1,12 +1,7 @@
 name: Release Helm Chart
 
 "on":
-  push:
-    tags:
-      # 2023.9.1, but not 2023.09.01
-      # 2023.11.5, but not 2023.11.05
-      # https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet
-      - '2[0-9][2-9][3-9].1?[0-9].[1-3]?[0-9]'
+  workflow_dispatch:
 
 jobs:
   release:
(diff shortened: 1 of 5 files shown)


.github/workflows/helm-release.yaml
```diff
@@ -147,3 +145,14 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ github.token }}
           RELEASE_VERSION: ${{ steps.get_version.outputs.RELEASE_VERSION }}
+
+      - name: Create Github Release + Tag
```
parkedwards, line 149: this release now doesn't really do anything other than document the changes reflected in this release

jamiezieziula (approved):
nice

PR #300, opened 2024-02-09
fix: add chart version timestamp in the patch slot, without hyphen

parkedwards:
so it will be:
```
2024.2.9125019

# instead of

2024.2.9-125019
```

as the [pre-release append](https://semver.org/#spec-item-9) is NOT treated as part of the normal versioning sequence:

> Pre-release versions have a **lower precedence than the associated normal version**. A pre-release version indicates that the version is unstable and might not satisfy the intended compatibility requirements

### DIFF
--- a/.github/workflows/helm-release.yaml
+++ b/.github/workflows/helm-release.yaml
@@ -21,7 +21,7 @@ jobs:
           # Enable pipefail so git command failures do not result in null versions downstream
           set -x
 
-          echo "RELEASE_VERSION=$(date +'%Y.%-m.%-d-%H%M%S')" >> $GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


.github/workflows/helm-release.yaml
```diff
@@ -21,7 +21,7 @@ jobs:
           # Enable pipefail so git command failures do not result in null versions downstream
           set -x
 
-          echo "RELEASE_VERSION=$(date +'%Y.%-m.%-d-%H%M%S')" >> $GITHUB_OUTPUT
+          echo "RELEASE_VERSION=$(date +'%Y.%-m.%-d%H%M%S')" >> $GITHUB_OUTPUT
```
parkedwards, line 24: the `%H%M%S` will **always** be 6 digits, so we can back into the "day" piece pretty consistently (6 chars from the right)

jamiezieziula (approved):

PR #317, opened 2024-04-22
Allow additional properties

### LINKED ISSUE #316: additionalProperties to true
We are currently trying to implement Prefect but need to set an additional property to get it properly working. I tried to create a PR for this but do not have the proper permissions.

Could you change `"additionalProperties": false` to `"additionalProperties": true`?

Thanks in advance!

fabrice-toussaint:
Closes #316

### DIFF
--- a/charts/prefect-server/values.schema.json
+++ b/charts/prefect-server/values.schema.json
@@ -38,7 +38,7 @@
       "title": "Server",
       "description": "server configuration",
       "form": true,
-      "additionalProperties": true,
+      "additionalProperties": false,
       "properties": {
         "image": {
           "type": "object",
@@ -145,7 +145,7 @@
           "title": "Autoscaling",
           "description": "server autoscaling configuration",
           "form": true,
-          "additionalProperties": true,
+          "additionalProperties": false,
           "properties": {
             "enabled": {
               "type": "boolean",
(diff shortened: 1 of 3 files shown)


jamiezieziula:
Hi @fabrice-toussaint, can you give us some more insight on the additional property that you would like to add and what it would be mapped to in your use case?

fabrice-toussaint:
@jamiezieziula we are trying to implement Prefect using vcluster to spin up a temporary environment. We want to be able to enable / disable specific charts (depending on the temporary environment). The other services that we use (i.e. postgres) allow us to set an additional property (i.e. enabled) so that we can include / exclude them.

charts/prefect-server/values.schema.json
```diff
@@ -1,7 +1,7 @@
 {
   "$schema": "http://json-schema.org/schema#",
   "type": "object",
-  "additionalProperties": false,
+  "additionalProperties": true,
```
jimid27, line 4: For context this means that something like:
Screenshot 2024-05-01 at 2 48 44 PM

is great.  But something like:
Screenshot 2024-05-01 at 2 49 21 PM

is still not permitted which makes sense for these use cases.

jimid27 (approved):
lgtm - appreciate you explaining your use case!

charts/prefect-server/values.schema.json
fabrice-toussaint: Yes exactly that would work perfectly for us

PR #318, opened 2024-04-24
prefect-server helm chart enhancement to support multi-port service

### LINKED ISSUE #314: Enhancing Prefect Server Helm Chart for Sidecar Containers
Currently, the chart supports sidecar containers by default through the extraContainers values specified in the values.yaml file. However, there’s a gap in the default behavior of the service itself.

The Prefect Server’s service exposes only a single service port (specifically, port 4200). While this suffices for basic functionality, we recognize that users often require additional features, such as logging, facilitated by sidecar containers. In a microservices architecture, it’s common to have multi-port services to accommodate these requirements.

To enhance the Prefect Server’s flexibility and usability, we propose modifying the Helm chart to create a multi-port service by default. This adjustment will allow users to seamlessly incorporate sidecar containers without manual configuration.

I plan to address this issue by submitting a pull request in the coming days.

baisystems:
prefect-server helm chart enhancement to support multi-port service.

- service template creates service name and ports based on the ports array from values.yaml file
- update deployment template to use the server port extracted from the ports array
- add a service port to ingress template and values
- updated helm chart validation schema

closes https://github.com/PrefectHQ/prefect-helm/issues/314


jamiezieziula:
Hi @baisystems, thanks for the PR! Is there a reason we can't accomplish this functionality by allowing for `additionalPorts` to be passed as an optional array[dictionary] and leaving the current handling of the required service ports as is?

baisystems:
Hi @jamiezieziula, there is no `additionalPorts` currently. If I understand you correctly, you are proposing to add a new `extraPorts` field under service in values.yaml file and use it like this in service template:

```
ports:
    - name: server-svc-port
      port: {{ .Values.service.port }}
      ...
    {{- range .Values.service.extraPorts }}
    - name: {{ .name }}
      port: {{ .port }}
      ...
    {{- end }}
```

If this is the case, my answer is `Yes`, we can accomplish this functionality with `extraPorts`. I do see the merits of your proposal: minimal changes to current code base as I don't need to change `deployment.yaml` and I don't need to add a new function `tartget-port.yaml`.

Your proposal also agree well with your current helm chart pattern `extra ...`, for example, `extraContainers` for server pod and `extraPaths` for server ingress.

I actually like your idea and can tailor my pr to implement the `extraPorts` field as I illustrated above. 

The second feature of this PR is its introduction of a new field `servicePort` to ingress. This will make the ingress service port configurable to accommodate the case where the ingress is serving a port other than `server-svc-port`. What do you think about this change?

jamiezieziula:
@baisystems thanks for the feedback! I look forward to reviewing your changes :)

I like the idea of the customizability of the ingress service port!

charts/prefect-server/values.yaml
```diff
@@ -185,17 +185,29 @@ serviceAccount:
...
   # -- service port
   port: 4200
   # -- target port of the server pod; also sets PREFECT_SERVER_API_PORT
   targetPort: 4200
-  # -- service Cluster IP
-  clusterIP: ""
   # -- service port if defining service as type nodeport
   nodePort: ""
 
+  extraPorts:
```
jamiezieziula, line 195:
```suggestion
  extraPorts: []
```

charts/prefect-server/values.yaml
```diff
@@ -185,17 +185,29 @@ serviceAccount:
...
+    # - name: sample-svc-port
+    #   # -- service port
+    #   port: 8080
+    #   # -- target port
+    #   targetPort: 8080
+    #   # -- service port if defining service as type nodeport
+    #   nodePort: ""
+      
+  # -- service type
+  type: ClusterIP
```
jamiezieziula, line 206:
```suggestion
```

charts/prefect-server/templates/service.yaml
```diff
@@ -26,6 +26,7 @@ spec:
   {{- if eq .Values.service.type "NodePort" }}
   externalTrafficPolicy: {{ .Values.service.externalTrafficPolicy | quote }}
   {{- end }}
+  {{- $serviceType := .Values.service.type }}
```
jamiezieziula, line 29: I don't think this is neccessary?

jamiezieziula (changes requested):
A few minor comments - thanks!

charts/prefect-server/values.yaml
baisystems: I will fix this in the next commit.

charts/prefect-server/templates/service.yaml
baisystems: I need to define this variable to hold the value of serviceType. The variable is used inside the loop of `{{- range .Values.service.extraPorts }}` to set nodePort. While inside the loop, it can't access the .Values.service.type due to context being changed. Is there better way to handle this situation?

jamiezieziula: ah i see - yes i think this makes sense!

baisystems:
@jamiezieziula I pushed a new commit which default the extraPorts value to []. Any other issue for this pr?

jamiezieziula:
@baisystems  - can you install precommit and run it? linters are failing.

baisystems:
@jamiezieziula I pushed a new commit. All the lint errors are fixed and the build is now passed. Could you please take another look?

charts/prefect-server/templates/service.yaml
mitchnielsen: You should be able to access the value with `$.Values.service.type` (notice the `$.` instead of just `.`).

There's an example at the bottom of https://helm.sh/docs/chart_template_guide/variables.

jamiezieziula:
Hi @baisystems - thanks! All linting looks healthy now. Sorry to do this, but one last thing - see @mitchnielsen's comment [above](https://github.com/PrefectHQ/prefect-helm/pull/318#discussion_r1608550571).

charts/prefect-server/templates/service.yaml
baisystems: @mitchnielsen  Thank you for the tip. It works. $ - this variable will always point to the root context. Great to learn this.

baisystems:
@jamiezieziula I tested @mitchnielsen's suggestion, helm install succeed with it. I will push an update commit soon.

baisystems:
@jamiezieziula I pushed a new commit, the 5th. Please review again.

jamiezieziula (approved):

PR #337, opened 2024-05-30
chore: skip chart releases for upstream release candidate builds

parkedwards:


.github/workflows/helm-release.yaml
```diff
@@ -28,6 +28,11 @@ jobs:
             https://github.com/PrefectHQ/prefect.git '[!prefect-]*.*.*' | tail -n1 | sed 's/.*\///' \
           )" >> $GITHUB_OUTPUT
 
+          if [[ $PREFECT_VERSION == 3.0rc* ]]; then
```
jimid27, line 31: I dont know if this is true but do we need to do anything for regex here? Like =~ 3.0rc*?

jimid27 (approved):

.github/workflows/helm-release.yaml
parkedwards: with double brackets `if [[ ]]`, you can do a pattern match with `*`

https://tldp.org/LDP/abs/html/comparison-ops.html

image

jimid27: sick TIL

PR #341, opened 2024-05-31
Change postgresql.useSubChart to .enabled

### LINKED ISSUE #336: Postgresql pod is still deployed even when setting `postgresql.enabled=false`
a postgresql pod, etc. is still deployed when "disabling" postgresql. You can reproduce by running

```bash
helm install prefect-server prefect/prefect-server --set postgresql.enabled=false
```

To block postgres from deploying, you would need to also set `postgresql.useSubChart=false`. This is confusing behavior. We should either better document this pattern or change the way in which we control the postgresql implementation

```bash
helm install prefect-server prefect/prefect-server --set postgresql.enabled=false --set postgresql.useSubChart=false
```
 
Relevant slack link [here](https://prefect-community.slack.com/archives/CL09KU1K7/p1717025127801229)

mitchnielsen:
## Summary

Changes `postgresql.useSubChart` to `postgresql.enabled`, following the common convention mentioned in https://helm.sh/docs/topics/charts/https://helm.sh/docs/topics/charts/.

This should hopefully make it clearer what that key does - and this key already existed, so by removing `useSubChart`, there's only one key to manipulate.

Other commits in this PR have commit messages explaining what and why.

Closes https://github.com/PrefectHQ/prefect-helm/issues/336

### Internal PostgreSQL enabled (default)

(Default values used)

```
$ helm template test . -f test.values.yaml --show-only charts/postgresql/templates/primary/statefulset.yaml | yq '.metadata.name'
test-postgresql
```

### Internal PostgreSQL disabled

```yaml
postgresql:
  enabled: false
```

```
$ helm template test . -f test.values.yaml --show-only charts/postgresql/templates/primary/statefulset.yaml | yq '.metadata.name'
Error: could not find template charts/postgresql/templates/primary/statefulset.yaml in chart
null
```

### Internal PostgreSQL + default Secret

(Default values used)

```
$ helm template test . -f test.values.yaml --show-only templates/deployment.yaml| yq '.spec.template.spec.containers[0].env[] | select(.name == "PREFECT_API_DATABASE_CONNECTION_URL") | .valueFrom.secretKeyRef.name'
prefect-server-postgresql-connection
```

### Internal PostgreSQL + existing Secret

```yaml
postgresql:
  enabled: true
  auth:
    existingSecret: my-external-pg-secret

```

```
$ helm template test . -f test.values.yaml --show-only templates/deployment.yaml| yq '.spec.template.spec.containers[0].env[] | select(.name == "PREFECT_API_DATABASE_CONNECTION_URL") | .valueFrom.secretKeyRef.name'
my-external-pg-secret
```

### External PostgreSQL + existing Secret

```yaml
postgresql:
  enabled: false
  auth:
    existingSecret: my-external-pg-secret
```

```
$ helm template test . -f test.values.yaml --show-only templates/deployment.yaml| yq '.spec.template.spec.containers[0].env[] | select(.name == "PREFECT_API_DATABASE_CONNECTION_URL") | .valueFrom.secretKeyRef.name'
my-external-pg-secret
```

### DIFF
--- a/charts/prefect-server/values.schema.json
+++ b/charts/prefect-server/values.schema.json
@@ -660,12 +660,6 @@
           "description": "external hostname for the postgresql service",
           "form": true
         },
-        "enabled": {
-          "type": "boolean",
-          "title": "Enabled",
-          "description": "enable use of bitnami/postgresql subchart",
-          "form": true
-        },
         "primary": {
           "type": "object",
           "title": "Primary",
--- a/charts/prefect-server/values.schema.json
+++ b/charts/prefect-server/values.schema.json
@@ -660,9 +660,9 @@
           "description": "external hostname for the postgresql service",
           "form": true
         },
-        "useSubChart": {
+        "enabled": {
           "type": "boolean",
-          "title": "Use Sub Chart",
+          "title": "Enabled",
           "description": "enable use of bitnami/postgresql subchart",
           "form": true
         },
--- a/charts/prefect-server/values.schema.json
+++ b/charts/prefect-server/values.schema.json
@@ -594,49 +594,41 @@
       "type": "object",
       "title": "PostgreSQL",
       "description": "Postgresql configuration",
-      "form": true,
       "properties": {
         "enabled": {
           "type": "boolean",
           "title": "Enabled",
-          "description": "enable postgresql",
-          "form": true
+          "description": "enable postgresql"
         },
         "auth": {
           "type": "object",
           "title": "Auth",
           "description": "postgresql authentication configuration",
-          "form": true,
           "properties": {
             "enablePostgresUser": {
               "type": "boolean",
               "title": "Enable Postgres User",
-              "description": "determines whether an admin user is created within postgres",
-              "form": true
+              "description": "determines whether an admin user is created within postgres"
             },
             "database": {
               "type": "string",
               "title": "Database",
-              "description": "name for a custom database",
-              "form": true
+              "description": "name for a custom database"
             },
             "username": {
               "type": "string",
               "title": "Username",
-              "description": "name for a custom user",
-              "form": true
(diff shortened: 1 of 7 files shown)


mitchnielsen:
Jotting down action items for myself:

Caught up with @jamiezieziula on this. The `postgresql.auth` key in `values.yaml` is Prefect-specific, so we need to allow `auth` settings to be set outside of the scope of `postgresql` when the built-in PG chart isn't used (`postgresql.enabled=false`).

We'll also want to support passing no auth and no database for those who want to test with the SQLite DB.

charts/prefect-server/values.schema.json
```diff
@@ -660,9 +660,9 @@
           "description": "external hostname for the postgresql service",
           "form": true
         },
-        "useSubChart": {
+        "enabled": {
```
jamiezieziula, line 663: this already exists on line 599

jamiezieziula (changes requested):
There is a duplicate `enabled` map in the json schema. Also, can you remove all `form: true` fields in the schema, these aren't needed (we removed most before but must have missed these)

charts/prefect-server/values.schema.json
mitchnielsen: Good catch, removed in 351e405.

mitchnielsen:
Removed `form` fields from the server schema in db29dcf.

jamiezieziula (approved):

parkedwards (approved):

mitchnielsen:
- Failures seem to be from hitting the Docker pull limit.
- Will proceed with merge soon but wanted to give the community member a chance to chime in.
- Will also consider if this could be a breaking change 🤔

pythonking6:
This is excellent!!!

PR #344, opened 2024-06-04
fix(release): skip upstream rc builds

parkedwards:
```
git ls-remote --tags --refs --sort="v:refname" https://github.com/PrefectHQ/prefect.git '[!prefect-]*.*.[!rc]'| tail -n1 | sed 's/.*\///'
2.19.4
```

this way, when 3.0 is out of RC, we don't have to do anything and can just continue releasing new chart versions


mitchnielsen:
🫠 

> 429 Too Many Requests - Server message: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit

mitchnielsen (approved):
Was able to get the same result from the `git ls-remote` command.

jamiezieziula (approved):

PR #354, opened 2024-07-08
feat(charts): support securityContext.capabilities

### LINKED ISSUE #352: Unable to configure securityContext.capabilities
It is best practice to drop all capabilities and only add those required. In our clusters we enforce this with policies so I want to add this to my values:
```
server:
  containerSecurityContext:
      runAsUser: 1001
      runAsNonRoot: true
      readOnlyRootFilesystem: true
      allowPrivilegeEscalation: false
      capabilities:
        drop:
          - ALL
 ```
This fails due to the schema validation which does not appear to be skippable with helm. Can the capabilities node be added to schema in the server and worker helm charts?

### PR DESCRIPTION
**Pull request by mitchnielsen**
## Summary

Adds support for setting `securityContext.capabilities on the server and worker charts.

Closes https://github.com/PrefectHQ/prefect-helm/issues/352

### prefect-server chart

Set up values to test:

```yaml
server:
  containerSecurityContext:
    capabilities:
      drop:
        - ALL

worker:
  containerSecurityContext:
    capabilities:
      drop:
        - ALL
  config:
    workPool: foo
  cloudApiConfig:
    accountId: bar
    workspaceId: baz

agent:
  containerSecurityContext:
    capabilities:
      drop:
        - ALL
```

Test the output:

```
$ for chart in $(ls charts | grep '^prefect-'); do helm template test charts/$chart -f test.values.yaml | yq '.spec.template.spec.containers[].securityContext.capabilities'; done

drop:
  - ALL
drop:
  - ALL
---
drop:
  - ALL
drop:
  - ALL
```

### DIFF
--- a/charts/prefect-agent/templates/deployment.yaml
+++ b/charts/prefect-agent/templates/deployment.yaml
@@ -122,8 +122,8 @@ spec:
           {{- if .Values.agent.resources }}
           resources: {{- toYaml .Values.agent.resources | nindent 12 }}
           {{- end }}
-          {{- if .Values.agent.containerSecurityContext }}
-          securityContext: {{- .Values.agent.containerSecurityContext | toYaml | nindent 12 }}
+          {{- with .Values.agent.containerSecurityContext }}
+          securityContext: {{- toYaml . | nindent 12 }}
           {{- end }}
           volumeMounts:
             - mountPath: /home/prefect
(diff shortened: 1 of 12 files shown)


charts/prefect-agent/templates/deployment.yaml
```diff
@@ -122,8 +122,8 @@ spec:
           {{- if .Values.agent.resources }}
           resources: {{- toYaml .Values.agent.resources | nindent 12 }}
           {{- end }}
-          {{- if .Values.agent.containerSecurityContext }}
-          securityContext: {{- .Values.agent.containerSecurityContext | toYaml | nindent 12 }}
+          {{- with .Values.agent.containerSecurityContext }}
```
parkedwards, line 125: 🚀

parkedwards (approved):

Issue #345, opened 2024-06-05
Support SQLite database


mitchnielsen:
## Summary

In https://github.com/PrefectHQ/prefect-helm/issues/336, we found that the database configuration was centralized within PostgreSQL - meaning database configuration couldn't be provided outside of the `postgresql` key, which matters if someone wants to use MySQL or even SQLite.

Let's consider providing a new top-level key for configuring the database that's considered if `postgresql.enabled=false`.

## Notes

The [Prefect docs](https://docs-3.prefect.io/3.0rc/manage/self-host) mention: 

> Prefect supports the following databases:
> 1. SQLite (default in Prefect): Recommended for lightweight, single-server deployments. SQLite requires essentially no setup.
> 2. PostgreSQL: Best for connecting to external databases, but requires additional setup (such as Docker). Prefect uses the [pg_trgm](https://www.postgresql.org/docs/current/pgtrgm.html) extension, so it must be installed and enabled.

Elsewhere, it says:

> Use SQLite database for testing only
> SQLite does not support multiprocessing. For high orchestration volume, use PostgreSQL.

So realisticically, most users should be on PostgreSQL. But the documentation shows you how to use SQLite, so I think it's reasonable to support that configuration in the charts. As mentioned in https://github.com/PrefectHQ/prefect-helm/pull/365, users are already using SQLite with the charts with some workarounds.

## Acceptance criteria

- [x] Database configuration can be provided outside of the top-level `postgresql` key if PostgreSQL is not in use.
- [x] SQLite specifically is tested to confirm nothing is hard-coded to work with PostgreSQL

mitchnielsen:
I bet this was indirectly addressed by [https://github.com/PrefectHQ/prefect-helm/pull/365](https://github.com/PrefectHQ/prefect-helm/pull/365/files), but let's try it to be sure and add an SQLite test case to be sure.

mitchnielsen:
@tmyhu - wanted to tag you since you mentioned you were using SQLite. We added direct support for it in https://github.com/PrefectHQ/prefect-helm/releases/tag/2024.12.5195120. If you have any suggestions feel free to create an issue/PR 🤝

tmyhu:
Awesome, that looks great. Thanks for letting me know @mitchnielsen

PR #365, opened 2024-07-23
fix(prefect-server): better support for internal and external database configs

### LINKED ISSUE #358: Changes to chart with `postgresql.enabled = false` and external postgres break deployment
Changes from #341 are breaking my deployment.

Context:  I use [movetokube/postgres-operator: Postgres operator for Kubernetes](https://github.com/movetokube/postgres-operator) to dynamically provision postgres as part of my gitops deployment process.  This operator creates unique auth info in a secret that _does not match the expected format_ of the prefect helm chart.

Prior to that change, I set:

```yaml
  valuesFrom:
    - kind: Secret
      name: &db_secret database-prefect
      valuesKey: DATABASE_NAME
      targetPath: postgresql.auth.database
    - kind: Secret
      name: *db_secret
      valuesKey: LOGIN
      targetPath: postgresql.auth.username
    - kind: Secret
      name: *db_secret
      valuesKey: PASSWORD
      targetPath: postgresql.auth.password
    - kind: Secret
      name: *db_secret
      valuesKey: HOST
      targetPath: postgresql.externalHostname
values:
  ...
  postgresql:
      enabled: true
      useSubChart: false
```

And the helpers picked up the values I was injecting in from the operator's secret at the valid targetPaths.

After #341, this approach does not work because the helm chart _expects_ that there will be a secret named `prefect-server-postgresql-connection` and does not create the secret as it used to based on the injected values.

I'd really appreciate it if the chart supported the use of external postgres while providing details in `postgres.auth` and `postgres.externalHostname` or supported passing the full `postgresql+asyncpg://{username}:{password}@{hostname}/{database}` connection url as an option.

The requirement to use the `prefect-server-postgresql-connection` secret that does not work with my gitops deployments!  Thank you!

mitchnielsen:
## Summary

**This will be a breaking change**, so we should release this in a new major version. The configuration keys have moved around. If we _really_ need to, we can try to build in some backwards-compatibility, but that always ends up looking pretty rough in Helm.

Ensures that the PostgreSQL Secret is created, even if `postgresql.enabled=false`. This ensures that we support a use case where folks want to use an external instance of PostgreSQL, but still want the Secret to automatically be generated with the proper connection string.

With a recent change, we would skip creation of this secret if PostgreSQL was disabled which forced users to create a Secret themselves. This change now allows them to continue providing the `auth` values and letting the chart build the Secret with the correct connection string.

I ended up needing to go a step further by removing any custom configuration that was injected under the `postgresql` key, which should be dedicated to overriding default values from the PostgreSQL chart. There's now a new `secret` key, which aligns with the other top-level keys (`deployment`, `service`, etc.). These values will be considered if `postgresql.enabled=false`.

Closes https://github.com/PrefectHQ/prefect-helm/issues/358

Follow-up to https://github.com/PrefectHQ/prefect-helm/pull/341.

## Testing

See the [added test suite](https://github.com/PrefectHQ/prefect-helm/blob/db-secret-config/charts/prefect-server/tests/database_test.yaml).


mitchnielsen:
@ahgraber @tmyhu @MRocholl - I _believe_ this should cover all the use cases you mentioned. Could you please give this a look when you have a moment and let me know if this needs any tweaking?

Trying to make it flexible enough to cover all use cases while keeping it as consistent/straightforward as possible. Thanks!

tmyhu:
@mitchnielsen it looks good to me, but won't make any difference for our use case since we are using SQLite. I assume that's not something you really want to support specifically since there is no hint about it anywhere in the helm chart I could see? We should probably migrate to PostgreSQL anyway, but for now I'm happy to continue creating the secret and connection string manually for SQLite.

ahgraber:
@mitchnielsen Thanks for this effort! 

The output connection string from the test deployment's secret is missing the `hostname` in the database address; will that still be pulled in from `postgresql.externalHostname`?  If so, I believe this covers the requested use case.

mitchnielsen:
> it looks good to me, but won't make any difference for our use case since we are using SQLite. I assume that's not something you really want to support specifically since there is no hint about it anywhere in the helm chart I could see? We should probably migrate to PostgreSQL anyway, but for now I'm happy to continue creating the secret and connection string manually for SQLite.

@tmyhu fair point. We have this in mind and are tracking in https://github.com/PrefectHQ/prefect-helm/issues/345. I'd like to provide the `auth` configuration outside of the context of the `postgresql` key to make the configuration more clear, and open the door for supporting SQLite since we provide that information in our self-hosted docs. That said, moving to PostgreSQL is probably the ideal longer-term move.

mitchnielsen:
@ahgraber great call-out, let me see if I can make a minimal change to make that configuration supported. If not, may need to jump ahead to providing this config outside of the `postgresql` key because of how some of the nested templates are used.

mitchnielsen:
@ahgraber @tmyhu @MRocholl - alrighty, bit of a rework here to make sure we have proper support for either internal or external database configuration. For some examples of the configuration you'd provide, check out the updated docs or the [unit test file](https://github.com/PrefectHQ/prefect-helm/blob/0f43bfc8f082caad1b2faf19a0ef74943ff345e4/charts/prefect-server/tests/database_test.yaml). Welcoming any feedback here 🤝

MRocholl:
@mitchnielsen sorry for the late reply. These changes should satisfy the requirements I believe 👍  Thx for the great work

mitchnielsen:
> sorry for the late reply. These changes should satisfy the requirements I believe 👍 Thx for the great work

@MRocholl - no worries at all, thank you for confirming 🤝

charts/prefect-server/values.yaml
```diff
@@ -340,8 +342,6 @@ postgresql:
     persistence:
       # -- enable PostgreSQL Primary data persistence using PVC
       enabled: false
-      # -- PVC Storage Request for PostgreSQL volume
-      size: 8Gi
```
parkedwards, line 344: does this default need to be re-implemented somewhere?

charts/prefect-server/values.yaml
```diff
@@ -293,7 +293,26 @@ ingress:
   ##           port:
   ##             name: http
 
-# Postgresql configuration
+# Secret configuration
+secret:
```
parkedwards, line 297: just so i understand, `.Values.secret` is specifically for BYO postgres credentials?

charts/prefect-server/values.yaml
mitchnielsen: Shouldn't need to, since it's [the default in Bitnami's values.yaml](https://github.com/bitnami/charts/blob/postgresql/12.12.10/bitnami/postgresql/values.yaml#L685). We could technically leave this here, but I deleted it to make it more clear that everything defined under `postgresql` is an override for the chart's defaults.

charts/prefect-server/values.yaml
mitchnielsen: Exactly, it gives us somewhere outside of the context of `postgresql.` to define database configs (so we can dedicate `postgresql.` settings for the PostgreSQL chart without overloading it with our own chart's configs).

parkedwards (approved):

jamiezieziula (approved):

PR #367, opened 2024-07-31
fix(docs): apply helm-docs fixes

mitchnielsen:
Applies a couple fixes related to `helm-docs` that were causing diffs in unrelated changes.


parkedwards (approved):

charts/prefect-worker/README.md.gotmpl
```diff
@@ -213,6 +213,14 @@ prefect work-pool get-default-base-job-template --type kubernetes > base-job-tem
 helm install prefect-worker prefect/prefect-worker -f values.yaml --set-file worker.config.baseJobTemplate.configuration=base-job-template.json
 ```
 
+#### Updating the Base Job Template
```
mitchnielsen, line 216: I think this being removed in https://github.com/PrefectHQ/prefect-helm/pull/368 was unintentional, so let's add it back for now. Can remove it later if we really don't want it.

PR #375, opened 2024-08-27
Add seccompProfile as valid podSecurityContext property

### LINKED ISSUE #374: Allow to set podSecurityContext seccompProfile
Hello,
and first of all thanks for supplying these charts.

For a deployment in a restricted environment I need to set the seccompProfile type but trying as such: 

````
server:
  podSecurityContext:
    seccompProfile:
      type: RuntimeDefault
````

results in

````
server.podSecurityContext: Additional property seccompProfile is not allowed
````
Could you make this possible? 
Thanks a lot!

### PR DESCRIPTION
**Pull request by dk-oc**
Looking at the chart i saw that only the allowed values prohibit setting seccompProfile. 

This would fix: #374

### CODE REVIEW

File: charts/prefect-server/values.schema.json [hunk 1]
```diff
@@ -313,6 +313,16 @@
               "type": "integer",
               "title": "FS Group",
               "description": "set server pod's security context fsGroup"
+            },
+            "seccompProfile": {
+              "type": "object",
```
parkedwards, line 318: is `additionalProperties` defaulting to true here (eg. if you need to pass `localhostProfile`)? or do we need to explicitly set it here?

jamiezieziula: Hi @dk-oc - just following up here.

jamiezieziula:
@dk-oc going to close this PR for now. Feel free to reopen in the future.

charts/prefect-server/values.schema.json
dk-oc: Sorry I missed the notification.
As far as I understand it defaults to true if a podSecurityContext is set. 
I use a values.yml as such and can deploy the chart in the restricted environment I am running
```
server:
  containerSecurityContext:
    capabilities:
      drop: ["ALL"]
  podSecurityContext:
    seccompProfile:
      type: RuntimeDefault
```

dk-oc:
> @dk-oc going to close this PR for now. Feel free to reopen in the future.

It seems I can not reopen this PR I could only create a new one.

mitchnielsen:
I reopened it for ya @dk-oc 👍🏼

dk-oc:
@mitchnielsen Thank you.
Is there anything missing in this PR which hinders a merging?

charts/prefect-server/values.schema.json
mitchnielsen: @dk-oc - I believe the question was more about what happens if you set `type: Localhost`, as that one requires an extra field to be set named `localhostProfile`:

```yaml
securityContext:
  seccompProfile:
    type: Localhost
    localhostProfile: my-profiles/profile-allow.json
```

- [K8s docs](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-seccomp-profile-for-a-container)
- [API docs](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.31/#seccompprofile-v1-core)

We'd need to either set `additionalProperties` here, or explicitly define the `localhostProfile`

mitchnielsen:
Thanks for checking in @dk-oc, I missed the added commits. I took a look and left a comment in the thread above.

mitchnielsen:
@dk-oc, I'm going to close this one for now. Reach out if you'd like us to reopen it.

PR #377, opened 2024-09-03
Bump default image versions for `prefect-worker` and `prefect-server` to 3

jamiezieziula:
Relates to PLA-228


charts/prefect-worker/values.yaml
```diff
@@ -59,7 +59,7 @@ worker:
     ## prefect tag is pinned to the latest available image tag at packaging time.  Update the value here to
     ## override pinned tag
     # -- prefect image tag (immutable tags are recommended)
-    prefectTag: 2-python3.11-kubernetes
+    prefectTag: 3-latest-kubernetes
```
parkedwards, line 62: dumb q, do we need to specify a python version?

jamiezieziula: hm maybe we should, otherwise it defaults to the latest python version (3.12.5)
```
ENV PYTHON_VERSION=3.12.5
```

jamiezieziula: reverted to 3.11

mitchnielsen (approved):
Looks good 👍🏼

mitchnielsen:
Might be handy to link to https://docs.prefect.io/3.0/resources/upgrade-to-prefect-3 in the release notes.

PR #378, opened 2024-09-03
chore: Remove `prefect-agent` chart

jamiezieziula:
Prefect `agents` have been fully deprecated with the GA of prefect 3.0.0, correspondingly we will remove support for the chart.

Resolves PLA-236


mitchnielsen (approved):
Not seeing any dangling references to agents. Looks good 👍🏼

PR #382, opened 2024-09-04
chore: remove the `rc` provision in the helm release workflow now that prefect 3 is GA

jamiezieziula:
Resolves PLA-237

### DIFF
--- a/.github/workflows/helm-release.yaml
+++ b/.github/workflows/helm-release.yaml
@@ -25,7 +25,7 @@ jobs:
           echo "RELEASE_VERSION=$(date +'%Y.%-m.%-d%H%M%S')" >> $GITHUB_OUTPUT
           echo "PREFECT_VERSION=$(\
             git ls-remote --tags --refs --sort="v:refname" \
-            https://github.com/PrefectHQ/prefect.git '[3].*.*[!rc]' | tail -n1 | sed 's/.*\///' \
+            https://github.com/PrefectHQ/prefect.git '[3].*.[!rc]' | tail -n1 | sed 's/.*\///' \
           )" >> $GITHUB_OUTPUT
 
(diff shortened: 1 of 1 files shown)


jimid27 (approved):

.github/workflows/helm-release.yaml
```diff
@@ -25,7 +25,7 @@ jobs:
           echo "RELEASE_VERSION=$(date +'%Y.%-m.%-d%H%M%S')" >> $GITHUB_OUTPUT
           echo "PREFECT_VERSION=$(\
             git ls-remote --tags --refs --sort="v:refname" \
-            https://github.com/PrefectHQ/prefect.git '[!prefect-]*.*.[!rc]' | tail -n1 | sed 's/.*\///' \
+            https://github.com/PrefectHQ/prefect.git '[!prefect-]*.*.*' | tail -n1 | sed 's/.*\///' \
```
parkedwards, line 28: these will include rcs - wonder if we should do something like this?

```
➜ git ls-remote --tags --refs --sort="v:refname" \
            https://github.com/PrefectHQ/prefect.git '[3]*.*.[!rc]'

c40d069de2d182af586eaf7909ee77beba17fd46	refs/tags/3.0.0
```

jamiezieziula: we aren't cutting any more RCs, but I guess its fine to leave as is?

parkedwards: >we aren't cutting any more RCs, but I guess its fine to leave as is?

right, the `[!rc]` just prevents them from being returned in the tag query, so we never run the risk of bundling an rc tag here

jamiezieziula: i see, ill update!

parkedwards (approved):

PR #384, opened 2024-09-08
All Charts: add configurable revision history limit for deployments

jamiezieziula:
This should result in a no-op for chart users, as the kubernetes default is 10. This will allow users to override this default value to a limit of their choosing.

Resolves PLA-269


charts/prefect-server/README.md
```diff
@@ -227,3 +228,5 @@ prefect-server:
 | serviceAccount.create | bool | `true` | specifies whether a ServiceAccount should be created |
 | serviceAccount.name | string | `""` | the name of the ServiceAccount to use. if not set and create is true, a name is generated using the common.names.fullname template |
 
+----------------------------------------------
+Autogenerated from chart metadata using [helm-docs v1.14.2](https://github.com/norwoodj/helm-docs/releases/v1.14.2)
```
mitchnielsen, line 232: Wonder why `helm-docs` only includes this sometimes. Btw, since this project has a `.mise.toml` you can do `mise install` and it'll adhere to the version we specify there.

mitchnielsen (approved):

PR #389, opened 2024-09-26
Add option for setting nodePort for LoadBalancers

gauthier-labs:
Add option for LoadBalancer nodePort


charts/prefect-server/templates/service.yaml
```diff
@@ -31,7 +31,7 @@ spec:
       port: {{ .Values.service.port }}
       protocol: TCP
       targetPort: {{ .Values.service.targetPort }}
-      {{- if and (eq .Values.service.type "NodePort") (not (empty .Values.service.nodePort)) }}
+      {{- if and (or (eq .Values.service.type "NodePort") (eq .Values.service.type "LoadBalancer")) (not (empty .Values.service.nodePort)) }}
```
mitchnielsen, line 34: Thanks for the contribution @gauthier-labs 🤝 

At first glance this looks good - rather than testing by hand, could you please add a test for it? You can see an example [here](https://github.com/PrefectHQ/prefect-helm/blob/main/charts/prefect-server/tests/database_test.yaml) for how we tested database settings. It uses [helm-unittest](https://github.com/helm-unittest/helm-unittest).

If not, no worries - we can address it some time soon.

mitchnielsen (changes requested):

charts/prefect-server/templates/service.yaml
gauthier-labs: Added test similar to NodePort for LoadBalancer type

mitchnielsen: Awesome, thank you!

PR #402, opened 2024-10-17
Prometheus Chart: Replace 'CHART_VERSION' with 'RELEASE_VERSION'

mitchnielsen:
For the Prometheus Chart release, use the correct variable name.

Related to https://linear.app/prefect/issue/PLA-400/cycle-4-catch-all


jimid27 (approved):
Did we never set a CHART_VERSION env variable?

jamiezieziula (approved):

PR #418, opened 2024-12-05
feat(prefect-server): support sqlite database

### LINKED ISSUE #345: Support SQLite database

mitchnielsen:
## Summary

Supports using SQLite for the database in place of PostgreSQL.

Closes https://linear.app/prefect/issue/PLA-139/support-sqlite-database

Closes https://github.com/PrefectHQ/prefect-helm/issues/345

## Testing

First, check out the the unit tests added for confirming general functionality.

For a functional test: 
1. Provide the documented values to enable SQLite, and deploy the `prefect-server` chart (confirming that no PostgreSQL Pod is deployed)
2. Access the application (port forward if not using an Ingress)
3. Create some data (such as a variable)
4. Delete the Server Pod
5. Reload the application and confirm data created earlier is still present


jamiezieziula (approved):
nice!!

PR #424, opened 2025-01-14
Support seccompProfile attribute

ondrejme:
This commit adds a seccompProfile attribute support for podSecurityContext of prefect-server. Defaults to kubernetes default value RuntimeDefault.

### DIFF
--- a/charts/prefect-server/values.schema.json
+++ b/charts/prefect-server/values.schema.json
@@ -318,6 +318,48 @@
               "type": "integer",
               "title": "FS Group",
               "description": "set server pod's security context fsGroup"
+            },
+            "seccompProfile": {
+              "type": "object",
+              "title": "Seccomp Profile",
+              "description": "Seccomp profile settings for the pod",
+              "properties": {
+                "type": {
+                  "type": "string",
+                  "enum": ["RuntimeDefault", "Unconfined", "Localhost"]
+                },
+                "localhostProfile": {
+                  "type": "string",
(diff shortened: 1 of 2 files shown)


charts/prefect-server/values.schema.json
```diff
@@ -318,6 +318,18 @@
...
               "description": "set server pod's security context fsGroup"
+            },
+            "seccompProfile": {
+              "type": "object",
+              "title": "Seccomp Profile",
+              "description": "Seccomp profile settings for the pod",
+              "properties": {
+                "type": {
+                  "type": "string",
+                  "enum": ["RuntimeDefault", "Unconfined", "Localhost"]
```
parkedwards, line 329: since there's a Localhost option here, so do we also need a localhostProfile string key?

https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#seccompprofile-v1-core

ondrejme: Good point, fixed. I wonder if I should add a comment about the _localhostProfile_ in case of using _Localhost_ type?

parkedwards: that would be great - maybe in the `values.yaml` key, so our auto-doc can document it in the values table

parkedwards (approved):

charts/prefect-server/values.schema.json
ondrejme: commend added - please check and feel free to merge if you agree

PR #429, opened 2025-01-22
Support seccompProfile attribute in the worker chart.

acarmel:
This commit adds a seccompProfile attribute support for podSecurityContext in the worker chart similarly to what is done in this PR https://github.com/PrefectHQ/prefect-helm/pull/424. Defaults to kubernetes default value RuntimeDefault.

### DIFF
--- a/charts/prefect-worker/values.yaml
+++ b/charts/prefect-worker/values.yaml
@@ -206,6 +206,11 @@ worker:
     runAsNonRoot: true
     # -- set worker pod's security context fsGroup
     fsGroup: 1001
+    # -- set worker pod's seccomp profile
+    seccompProfile:
+      type: RuntimeDefault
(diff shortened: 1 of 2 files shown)


parkedwards (approved):

jamiezieziula (changes requested):
Just a nit on the values comment

charts/prefect-worker/values.yaml
```diff
@@ -206,6 +206,11 @@ worker:
     runAsNonRoot: true
     # -- set worker pod's security context fsGroup
     fsGroup: 1001
+    # -- set server pod's seccomp profile
```
jamiezieziula, line 209:
```suggestion
    # -- set worker pod's seccomp profile
```

acarmel: fixed. thanks for catching that

jamiezieziula (approved):

PR #434, opened 2025-01-28
[worker] update chart to reference `self-managed` rather than `self-hosted` for Self Managed Cloud

jamiezieziula:
Also updates `mise` deps

Relates to PLA-935


prefectcboyd:
Looks good, tested on self-managed -
image

prefectcboyd (approved):
🔥

PR #435, opened 2025-01-29
Add workflow to update mise tool versions

mitchnielsen:
## Summary

Adds a workflow that updates the versions in .mise.toml and commits and pushes the changes.

Related to https://linear.app/prefect/issue/PLA-961/automate-updating-dependencies-defined-in-mise

## Testing

From what I can tell, I can't really run this until we merge it. This is a completely separate workflow with no dependencies or downstream calls, so we should be safe to merge this, run it, and then tweak as needed.


mitchnielsen:
Might move this into a reusable workflow

jamiezieziula (approved):

PR #437, opened 2025-01-31
fix(actions): adjust job name for mise workflow

mitchnielsen:
The job name was a copy pasta error.

Related to https://linear.app/prefect/issue/PLA-961/automate-updating-dependencies-defined-in-mise

### DIFF
--- a/.github/workflows/update-mise-tools.yaml
+++ b/.github/workflows/update-mise-tools.yaml
@@ -9,7 +9,7 @@ name: Update mise tool versions
 permissions: {}
 
 jobs:
-  update-mise-tools:
+  update_mise_tools:
     runs-on: ubuntu-latest
     permissions:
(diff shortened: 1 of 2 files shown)


.github/workflows/update-mise-tools.yaml
```diff
@@ -9,7 +9,7 @@ name: Update mise tool versions
 permissions: {}
 
 jobs:
-  updatecli_major:
+  update-mise-tools:
```
mitchnielsen, line 12:
```suggestion
  update_mise_tools:
```

Maybe we want underscores here? Forgot which one is the ideal/standard

jamiezieziula: yeah i think underscores!

jamiezieziula (approved):

PR #446, opened 2025-02-10
chore(docs): document how to set imagePullSecrets in baseJobTemplate

mitchnielsen:
## Summary

Documents how to set imagePullSecrets in the baseJobTemplate.

Will be expanded as part of https://linear.app/prefect/issue/PLA-1020/expand-documentation-on-how-to-use-basejobtemplate

Closes https://linear.app/prefect/issue/PLA-1029/document-how-to-set-image-pull-secrets-in-the-base-job-template

## Testing

I installed the `prefect-server` chart, and then deployed the `prefect-worker` chart with this config:

```yaml
worker:
  apiConfig: server
  serverApiConfig:
    apiUrl: http://prefect-server.prefect.svc.cluster.local:4200/api
  config:
    workPool: test-pool
    baseJobTemplate:
      configuration: |
        # ...
```

I then confirmed the configuration appeared as expected in the UI.

### DIFF
--- a/charts/prefect-worker/README.md
+++ b/charts/prefect-worker/README.md
@@ -226,8 +226,6 @@ The worker uses the [base job template](https://docs.prefect.io/v3/deploy/infras
 to create the Kubernetes job that executes your workflow. The base job template configuration can be modified by setting
 `worker.config.baseJobTemplate.configuration`.
 
-**Note**: if the target work pool (`config.workPool`) already exists, the base job template passed **will be ignored**.
-
 1. Define the base job template in a local file. To get a formatted template, run the following command & store locally in `base-job-template.json`:
 
 ```bash
--- a/charts/prefect-worker/README.md
+++ b/charts/prefect-worker/README.md
@@ -259,7 +259,11 @@ worker:
              "spec": {
                "template": {
                  "spec": {
-                   "imagePullSecrets": "my-pull-secret"
+                   "imagePullSecrets": [
+                     {
+                       "name": "my-pull-secret"
+                     }
+                   ]
                  }
                }
              }
--- a/charts/prefect-worker/README.md
+++ b/charts/prefect-worker/README.md
@@ -272,7 +272,7 @@ in the top right corner > `Edit` > `Base Job Template` section > `Advanced` tab.
 
 #### Updating the Base Job Template
 
-If a base job template is set through Helm (via either `.Values.worker.config.baseJobTemplate.configuration` or `.Values.worker.config.baseJobTemplate.existingConfigMapName`), we'll run an optional `initContainer` that will sync the template configuration to the work pool named in `.Values.worker.config.workPool`.
+If a base job template is set through Helm (via either `worker.config.baseJobTemplate.configuration` or `worker.config.baseJobTemplate.existingConfigMapName`), we'll run an optional `initContainer` that will sync the template configuration to the work pool named in `worker.config.workPool`.
 
 Any time the base job template is updated, the subsequent `initContainer` run will run `prefect work-pool update <work-pool-name> --base-job-template <template-json>` and sync this template to the API.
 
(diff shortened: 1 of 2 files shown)


charts/prefect-worker/README.md
```diff
@@ -241,17 +241,46 @@ prefect work-pool get-default-base-job-template --type kubernetes > base-job-tem
...
+  config:
+    baseJobTemplate:
+      configuration: |
+       {
+         "job_configuration": {
+           "job_manifest": {
+             "spec": {
+               "template": {
+                 "spec": {
+                   "imagePullSecrets": "my-pull-secret"
```
mitchnielsen, line 262:
```
$ k explain deployment.spec.template.spec.imagePullSecrets
GROUP:      apps
KIND:       Deployment
VERSION:    v1

FIELD: imagePullSecrets <[]LocalObjectReference>


DESCRIPTION:
    ImagePullSecrets is an optional list of references to secrets in the same
    namespace to use for pulling any of the images used by this PodSpec. If
    specified, these secrets will be passed to individual puller implementations
    for them to use. More info:
    https://kubernetes.io/docs/concepts/containers/images#specifying-imagepullsecrets-on-a-pod
    LocalObjectReference contains enough information to let you locate the
    referenced object inside the same namespace.

FIELDS:
  name  <string>
    Name of the referent. This field is effectively required, but due to
    backwards compatibility is allowed to be empty. Instances of this type with
    an empty value here are almost certainly wrong. More info:
    https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
```

charts/prefect-worker/README.md
```diff
@@ -241,17 +241,50 @@ prefect work-pool get-default-base-job-template --type kubernetes > base-job-tem
...
+             "spec": {
+               "template": {
+                 "spec": {
+                   "imagePullSecrets": [
+                     {
+                       "name": "my-pull-secret"
+                     }
+                   ]
+                 }
+               }
```
mitchnielsen, line 268:
```
       {
         "job_configuration": {
           "job_manifest": {
             "spec": {
               "template": {
                 "spec": {
                   "imagePullSecrets": [
                     {
                       "name": "my-pull-secret"
                     }
                   ]
                 }
               }
             }
           }
         }
       }
```

charts/prefect-worker/README.md
```diff
@@ -241,17 +241,50 @@ prefect work-pool get-default-base-job-template --type kubernetes > base-job-tem
...
+The worker uses the [base job template](https://docs.prefect.io/v3/deploy/infrastructure-concepts/work-pools#base-job-template)
+to create the Kubernetes job that executes your workflow. The base job template configuration can be modified by setting
+`worker.config.baseJobTemplate.configuration`. For example, to provide image pull secrets for the container running your flow,
+provide the following values:
+
+```yaml
+worker:
+  config:
+    baseJobTemplate:
+      configuration: |
```
jamiezieziula, line 255: quick question -- if you provide this and nothing else, does this merge with the base base job template? or will it overwrite it to _just_ contain this information?

mitchnielsen: Great question - I checked the [source code](https://github.com/PrefectHQ/prefect/blob/e38b549fa2ae4cd70c5a18bb0dc9cb34cf2ca1fa/src/prefect/cli/work_pool.py#L414-L415) and it looks like we don't merge it over the default, we just take whatever is given (using `prefect work-pool update`). Looks like this is a little different than providing overrides in `prefect.yaml`, for example.

As a test, I created a work pool with the default base job template added under `worker.config.baseJobTemplate`. Then I removed a field, in this case `restartPolicy`, and upgraded the chart. At that point, `restartPolicy` was no longer present. This confirms there's no merging with the base. I'll add a note 👍🏼

charts/prefect-worker/README.md
```diff
@@ -233,25 +233,63 @@ pip install prefect-kubernetes
 prefect work-pool get-default-base-job-template --type kubernetes > base-job-template.json
 ```
 
-2. Modify the base job template as needed
+2. Modify the base job template as needed. Keep in mind that modifications are not merged with the default template. The configuration
```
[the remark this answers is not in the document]
mitchnielsen: Fair point - I took some of that away when I realized there was a section above that gave higher-level context. I updated it there in a6bcffc.

charts/prefect-worker/README.md.gotmpl
```diff
@@ -222,9 +222,13 @@ serviceAccount:
 
 ### Configuring a Base Job Template on the Worker
 
-If you want to define the [base job template](https://docs.prefect.io/concepts/work-pools/#base-job-template) of the worker and pass it as a value in this chart, you will need to do the following. **Note** if the `workPool` already exists, the base job template passed **will** be ignored.
+The worker uses the [base job template](https://docs.prefect.io/v3/deploy/infrastructure-concepts/work-pools#base-job-template)
+to create the Kubernetes job that executes your workflow. The base job template configuration can be modified by setting
+`worker.config.baseJobTemplate.configuration`.
 
-1. Define the base job template in a local file. To get a formatted template, run the following command & store locally in `base-job-template.json`
+**Note**: if the target work pool (`config.workPool`) already exists, the base job template passed **will be ignored**.
```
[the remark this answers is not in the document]
jamiezieziula: hm good question. it _is_ true i think - but maybe with the addition of the init container it is no longer true for our helm charts.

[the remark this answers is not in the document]
mitchnielsen: I bet that's it, since we're directly using `prefect work-pool update <json-file>` now. I'll remove that warning.

jamiezieziula (approved):

PR #448, opened 2025-02-12
Add PREFECT_API_AUTH_STRING to init container

n0tlikethatplz:
Right now, the init container does not work with basic auth as it will be unable to authenticate.


jamiezieziula (approved):
thanks for the contribution!

PR #449, opened 2025-02-13
feat(prefect-server): add custom entrypoint and args

fatih-acar:
This can be useful if we want to start a custom image for the server that has a different entrypoint and/or args.

### DIFF
--- a/charts/prefect-server/values.schema.json
+++ b/charts/prefect-server/values.schema.json
@@ -200,6 +200,22 @@
             "type": "object"
           }
         },
+        "args": {
+          "type": "array",
+          "title": "Custom container command arguments",
+          "description": "array with arguments",
+          "items": {
+            "type": "string"
+          }
+        },
+        "command": {
+          "type": "array",
+          "title": "Custom container entrypoint",
+          "description": "array with container entrypoint",
+          "items": {
+            "type": "string"
+          }
+        },
         "revisionHistoryLimit": {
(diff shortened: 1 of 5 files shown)


mitchnielsen:
Thanks for the contribution @fatih-acar!

Two quick requests - could you please:
1. Add these new fields to the chart's `values.yaml` with comments (so they're generated into the README for documentation purposes)
2. Add unit tests for these new fields (in `server_test.yaml`) to confirm the changes work as expected

Reach out if we can clarify/help with any of that - thanks again 👍🏼

fatih-acar:
Thanks, I have added requested changes.

charts/prefect-server/values.schema.json
```diff
@@ -200,6 +200,22 @@
...
+          }
+        },
+	"command": {
+          "type": "array",
+          "title": "Custom container entrypoint",
+          "description": "array with container entrypoint",
+          "items": {
+            "type": "string"
+          }
+        },
```
mitchnielsen, line 218: I _think_ the alignment is a bit off here - the `args` and `command` keys themselves need to be indented to the right a bit.

fatih-acar: Good catch, there was a tab vs space issue... thank you.

mitchnielsen (approved):

mitchnielsen:
Thanks again @fatih-acar 🤝

PR #479, opened 2025-04-04
fix(prefect-server): update ingress.servicePort type from string to both string and integer

janaki-sasidhar:
Updates the JSON schema to correctly define the ingress.servicePort as an both string and integer type rather than only  string.

### Summary

Updates the JSON schema to correctly define the ingress.servicePort as an integer type rather than a string, aligning with its actual usage as a numeric port value.

### Requirements

### DIFF
--- a/charts/prefect-server/values.schema.json
+++ b/charts/prefect-server/values.schema.json
@@ -776,7 +776,7 @@
         "servicePort": {
           "type": ["integer", "string"],
           "title": "Service Port Name",
-          "description": "service port name"
+          "description": "service port name or number"
         },
         "className": {
           "type": "string",
--- a/charts/prefect-server/values.schema.json
+++ b/charts/prefect-server/values.schema.json
@@ -774,7 +774,7 @@
           "description": "enable ingress record generation for server"
         },
         "servicePort": {
-          "type": "string",
+          "type": "integer",
           "title": "Service Port Name",
           "description": "service port name"
         },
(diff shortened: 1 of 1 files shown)


charts/prefect-server/values.schema.json
```diff
@@ -774,7 +774,7 @@
           "description": "enable ingress record generation for server"
         },
         "servicePort": {
-          "type": "string",
+          "type": "integer",
```
mitchnielsen, line 777: I took a look at this, and we use a [Bitnami `common` chart helper](https://github.com/bitnami/charts/blob/d1ea2593fc3a34439b742ac077edec74a1b8b3a6/bitnami/common/templates/_ingress.tpl#L8-L34) to define the backend. There, you can see it has conditional logic to consider if the `servicePort` value is a string or an integer (a port name or a port number).

So we can actually allow both here (and update the description):

```json
        "servicePort": {
          "type": [
            "integer",
            "string",
          ],
          "title": "Service Port",
          "description": "service port name or number"
        },
```

charts/prefect-server/values.schema.json
```diff
@@ -774,7 +774,7 @@
           "description": "enable ingress record generation for server"
         },
         "servicePort": {
-          "type": "string",
+          "type": ["integer", "string"],
           "title": "Service Port Name",
           "description": "service port name"
```
mitchnielsen, line 779:
```suggestion
          "description": "service port name or number"
```

mitchnielsen (approved):

mitchnielsen:
Thank you @janaki-sasidhar 🤝

PR #517, opened 2025-08-12
feat(worker): allow namespace override for worker role

### LINKED ISSUE #516: Support dedicated namespaces for worker jobs?
Hello, I was wondering if it would make sense to allow a target namespace override: https://github.com/PrefectHQ/prefect-helm/blob/main/charts/prefect-worker/templates/role.yaml#L6

so potentially i can set up a separate namespace for running the jobs of a particular worker instead of having them run on the same namespace?

yushao2:
### Summary

There is a scenario where I would like my jobs to execute in a different namespace than the worker. As such I believe the role namespace could allow for overrides in order for use cases where users would like to segregate the namespaces where jobs execute and for the worker deployments

closes #516 

### Requirements

- [ ] Relevant labels are added
- [ ] `Draft` status is used until ready for review

### DIFF
--- a/charts/prefect-worker/values.yaml
+++ b/charts/prefect-worker/values.yaml
@@ -328,6 +328,7 @@ serviceAccount:
 role:
   # -- specifies whether a Role should be created
   create: true
+  # -- the namespace to deploy the role to. If not provided, deploys to the same namespace as the worker
   namespace: null
   ## List of extra role permissions
   ## e.g:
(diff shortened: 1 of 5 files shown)


charts/prefect-worker/values.yaml
```diff
@@ -328,6 +328,7 @@ serviceAccount:
 role:
   # -- specifies whether a Role should be created
   create: true
+  namespace: null
```
jamiezieziula, line 331:
```suggestion
  # -- the namespace to deploy the role to. If not provided, deploys to the same namespace as the worker
  namespace: null
```

jamiezieziula (approved):

PR #547, opened 2025-09-24
feat(exporter): more flexible configuration of env vars on deployment

yushao2:
### Summary

I have a use case to add an env var from a secret, but the current chart does not allow me to do so easily, hence I'm using the current mechanism that is available on prefect-worker charts to allow myself to do the same on the exporter charts.

I noticed that there is already an `env` variable, but it is insufficient for me to work with secrets. I left the existing logic in place to avoid a breaking change here

### Requirements

### DIFF
--- a/charts/prometheus-prefect-exporter/README.md
+++ b/charts/prometheus-prefect-exporter/README.md
@@ -95,6 +95,7 @@ basicAuth:
 | basicAuth.existingSecret | string | `""` | name of existing secret containing basic auth credentials. takes precedence over authString. must contain a key `auth-string` with the value of the auth string |
 | csrfAuth | bool | `false` | Enable CSRF authentication (Only set to true if Prefect Server has CSRF enabled) |
 | env | object | `{}` | Environment variables to configure application |
+| extraEnvVars | list | `[]` | List of advanced environment variable configs to add to exporter pods |
 | fullnameOverride | string | `""` | String to fully override common.names.fullname template |
 | image | object | `{"pullPolicy":"IfNotPresent","repository":"prefecthq/prometheus-prefect-exporter","tag":"1.1.0"}` | Image registry |
 | imagePullSecrets | list | `[]` | Global Docker registry secret names as an array |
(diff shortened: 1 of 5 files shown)


mitchnielsen:
Thanks for the contribution here @yushao2 🤝 

This makes sense to me - we need a way to expose environment variables with one of the other mechanisms.

My only suggestion is that we try to follow the pattern we use in the `prefect-server` and `prefect-worker` charts. For example, [here](https://github.com/PrefectHQ/prefect-helm/blob/05ae8e44ef9a574236d8b471b10e0a3b2963a332/charts/prefect-worker/templates/deployment.yaml#L290-L297) is how we configure environment variables from ConfigMaps or Secrets in the `prefect-worker` deployment.

On that note, though: honestly this pattern in those two charts of `extraEnv{type}` is pretty verbose. Your method has the benefit of allowing end users to specify any YAML they like, with the tradeoff of providing less guidance. If we stuck with your method, we'd want to make sure there was an example in the docs and/or in the values file.

Let me know what you think. Thanks!

yushao2:
> Thanks for the contribution here @yushao2 🤝
> 
> This makes sense to me - we need a way to expose environment variables with one of the other mechanisms.
> 
> My only suggestion is that we try to follow the pattern we use in the `prefect-server` and `prefect-worker` charts. For example, [here](https://github.com/PrefectHQ/prefect-helm/blob/05ae8e44ef9a574236d8b471b10e0a3b2963a332/charts/prefect-worker/templates/deployment.yaml#L290-L297) is how we configure environment variables from ConfigMaps or Secrets in the `prefect-worker` deployment.
> 
> On that note, though: honestly this pattern in those two charts of `extraEnv{type}` is pretty verbose. Your method has the benefit of allowing end users to specify any YAML they like, with the tradeoff of providing less guidance. If we stuck with your method, we'd want to make sure there was an example in the docs and/or in the values file.
> 
> Let me know what you think. Thanks!

@mitchnielsen ,

i thought about that, but using extraEnvSecrets would require you to store the env var name as the key in the secrets, which would not fulfill the needs of this. this takes inspiration from what's available in the worker chart's deployment template actually, so it is indeed something that is supported

yushao2:
@mitchnielsen added!

mitchnielsen:
🚀

mitchnielsen (approved):

mitchnielsen (approved):

PR #570, opened 2025-11-15
fix(prefect-worker,prefect-server): omit replica in deployment spec if autoscale is enabled

youngern:
### Summary

The presence of `.spec.replica` in `Deployment` causes issues with autoscaling when HPA is configured and applied via a Deployment Controller.

Reference: https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#migrating-deployments-and-statefulsets-to-horizontal-autoscaling

> When an HPA is enabled, it is recommended that the value of spec.replicas of the Deployment and / or StatefulSet be removed from their [manifest(s)](https://kubernetes.io/docs/reference/glossary/?all=true#term-manifest). If this isn't done, any time a change to that object is applied, for example via kubectl apply -f deployment.yaml, this will instruct Kubernetes to scale the current number of Pods to the value of the spec.replicas key. This may not be desired and could be troublesome when an HPA is active, resulting in thrashing or flapping behavior

Also updates some failing `charttest` linting rules.

### Requirements


mitchnielsen:
> The linting issue seems to be a problem running `charttest` on `main` and not mine. Any advice on how to proceed here?

Good find, thank you! I'm working on a fix for this in https://github.com/PrefectHQ/prefect-helm/pull/571. Once that's merged you should be able to pull in those changes (or rebase). This should also address and remove the unrelated changes to the readme that ended up in this PR from `pre-commit`. I'll follow up with you when that's ready.

mitchnielsen (approved):
Thanks for this @youngern, great catch

PR #583, opened 2025-12-18
fix(prefect-worker): correct jobNamespace parameter description

### LINKED ISSUE #582: Worker setting jobNamespace does not set the Workpool namespace option
When setting up a worker using the most recent helm,
i can set the jobNamespace in the yaml file

```
worker:
  apiConfig: selfHostedServer

  config:
    workPool: pool1
    name: worker1
    jobNamespace: bi-prefect-dummy
```

but this setting does not set the settings of the worker (on a clean new deployment of the helm)

Image
 
.
sure, i can modify the setting by hand then.

Notably, https://github.com/PrefectHQ/prefect-helm/blob/main/charts/prefect-worker/values.yaml states:

```
    # -- the namespace where the jobs would spawn. If unset, spawns in the same namespace as the worker controller. Requires role.namespace to be set with the same value.
    jobNamespace: null
```

so either there is a bug somewhere or the comment here is misleading, as the job will spawn in the namespace `default` and fail due to missing permissions for this namespace.

The role.namespace is set in the yaml as requested in the comment:

```
serviceAccount:
  create: true
  name: "prefect-worker-sa"

role:
  create: true
  namespace: bi-prefect-dummy
  extraPermissions:
    - apiGroups: [""]
      resources: ["pods", "services"]
      verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]

rolebinding:
  create: true
```

jamiezieziula:
## Summary

Corrects the documentation for the `worker.config.jobNamespace` parameter, which was incorrectly described as controlling where worker jobs are deployed.

The parameter actually configures `PREFECT_INTEGRATIONS_KUBERNETES_OBSERVER_NAMESPACES`, which tells the Kubernetes observer which namespace(s) to watch for pod events and job crash detection.

## Changes

- Updated `values.yaml` description to clarify the parameter configures observer namespace watching
- Updated `README.md` with the same corrected description
- Added guidance that job deployment namespace is configured via the work pool's base job template

Resolves #582

### DIFF
--- a/charts/prefect-server/README.md
+++ b/charts/prefect-server/README.md
@@ -328,6 +328,7 @@ the HorizontalPodAutoscaler.
 | backgroundServices.podSecurityContext.runAsNonRoot | bool | `true` | set background-services pod's security context runAsNonRoot |
 | backgroundServices.podSecurityContext.runAsUser | int | `1001` | set background-services pod's security context runAsUser |
 | backgroundServices.priorityClassName | string | `""` | priority class name to use for the background-services pods; if the priority class is empty or doesn't exist, the background-services pods are scheduled without a priority class |
+| backgroundServices.replicaCount | int | `1` | number of background-services replicas to deploy |
 | backgroundServices.resources.limits | object | `{"cpu":"1","memory":"1Gi"}` | the requested limits for the background-services container |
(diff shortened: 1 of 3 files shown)


mitchnielsen (approved):
nice one

desertaxle (approved):

PR #603, opened 2026-03-05
fix(prefect-server): add ephemeral storage requests and limits

lostkamp:
### Summary

Hi,

sometimes it is necessary to specify these, so I thought I'd add them. I added them as comments to the values.yaml like they are already in other sections there.

### Requirements

- [ ] Body includes `Closes <issue>`, if available

### DIFF
--- a/charts/prefect-server/values.yaml
+++ b/charts/prefect-server/values.yaml
@@ -237,9 +237,11 @@ migrations:
     requests:
       cpu: 100m
       memory: 128Mi
+      # ephemeral-storage:
     limits:
       cpu: 500m
       memory: 256Mi
+      # ephemeral-storage:
   # -- job security context configuration
   securityContext:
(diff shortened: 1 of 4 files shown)


charts/prefect-server/values.yaml
mitchnielsen: Could we get some unit tests in place for these new potential values? You can see [lines 248-277 of server_test.yaml](https://github.com/PrefectHQ/prefect-helm/blob/bada6dc66ff92050ed5e5c2064b0ed411d12a209/charts/prefect-server/tests/server_test.yaml#L241-L277) for an example.

mitchnielsen (changes requested):
Thanks for the contribution @lostkamp! Just two small requests.

mitchnielsen (approved):
Thanks again @lostkamp 🤝

PR #606, opened 2026-03-06
fix(server): set HOME and podSecuritycontext for migration job

### LINKED ISSUE #605: Running migrations in job does not work with ssl enabled
[The example provided in the docs](https://github.com/PrefectHQ/prefect-helm/tree/main/charts/prefect-server#connecting-with-ssl-configured) unfortunately does not work. 

For the example to work: 
1. the `HOME` env variable should be set on the migrations job, otherwise asyncpg will look for `.postgresql` at the root folder. 
2. the `PodSecurityContext` should be set as in the `backgroundServices` section (**especially** the `fsGroup` property)

Unfortunately, there is no solution through the helm chart currently: 
1. It's not possible to set fsGroup
2. I also cannot run migrations on start of the server process, and run backgroundServices separately (which would also solve the issue).

mathijscarlu:
### Summary

Closes https://github.com/PrefectHQ/prefect-helm/issues/605

This change adds the following to the separate migrations job: 
* HOME env variable
* `podSecurityContext`, as specified in the `backgroundServices` section

### Requirements

### DIFF
--- a/charts/prefect-server/README.md
+++ b/charts/prefect-server/README.md
@@ -408,6 +408,9 @@ the HorizontalPodAutoscaler.
 | migrations.extraVolumeMounts | list | `[]` | additional volume mounts for the migration job |
 | migrations.extraVolumes | list | `[]` | additional volumes for the migration job |
 | migrations.nodeSelector | object | `{}` | node labels for migration job pods assignment |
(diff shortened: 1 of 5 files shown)


mitchnielsen (changes requested):
Left a comment requesting small changes 👍🏼

mitchnielsen (approved):

PR #615, opened 2026-04-02
feat(prefect-server): add support for PREFECT_SERVER_DOCKET_URL

### LINKED ISSUE #613: Add support for setting `PREFECT_SERVER_DOCKET_URL`
[Prefect docs](https://docs.prefect.io/v3/advanced/self-hosted#redis-setup) recommend setting up Docket to use Redis for large deployments. It'd be awesome if we could do this using the helm chart, preferably with support for reading from k8 secrets, instead of having to use the env variable!

jamiezieziula:
## Summary

- Add first-class Helm values for configuring `PREFECT_SERVER_DOCKET_URL`, which Prefect uses to coordinate background services (scheduler, late run detection, automation triggers)
- Support reading the Docket URL from an existing Kubernetes secret via `existingSecret` / `existingSecretKey`, following the same pattern as Redis password secret support
- Backwards compatible: when no docket values are configured (the default), no env var is injected and Prefect falls back to its built-in `memory://` backend

### Configuration Options

```yaml
backgroundServices:
  messaging:
    docket:
      # Option 1: plain-text URL
      url: "redis://redis-host:6379/1"

      # Option 2: from a Kubernetes secret (takes precedence over url)
      existingSecret: "my-docket-secret"
      existingSecretKey: "docket-url"  # default key name
```

Closes #613

## Test plan

- [x] `helm template` renders correctly with no docket config (env var absent)
- [x] `helm template` renders correctly with `docket.url` set (plain value)
- [x] `helm template` renders correctly with `docket.existingSecret` set (secretKeyRef)
- [x] `helm template` renders correctly with custom `existingSecretKey`
- [x] `existingSecret` takes precedence over `url` when both are set
- [x] Env var appears in both server and background-services deployments

### DIFF
--- a/charts/prefect-server/values.yaml
+++ b/charts/prefect-server/values.yaml
@@ -454,9 +454,9 @@ backgroundServices:
       # leave empty to use the default in-memory backend (memory://)
       url: ""
 
-      # -- name of an existing Kubernetes secret containing the Docket URL
-      # takes precedence over the url field above
-      # the secret should contain a key with the full Redis URL including any credentials
+      # -- name of an existing Kubernetes secret containing the Docket URL.
+      # Takes precedence over the url field above.
+      # The secret should contain a key with the full Redis URL including any credentials.
       existingSecret: ""
 
       # -- key within the existing secret that contains the Docket URL
--- a/charts/prefect-server/values.yaml
+++ b/charts/prefect-server/values.yaml
@@ -449,9 +449,9 @@ backgroundServices:
     # which only works for single-server deployments.
     # For high-availability deployments, configure Docket to use Redis.
     docket:
-      # -- URL for the Docket scheduler backend
-      # examples: redis://host:6379/0, redis://host:6379/0 (TLS), redis://user:pass@host:6379/0
-      # leave empty to use the default in-memory backend (memory://)
+      # -- URL for the Docket scheduler backend.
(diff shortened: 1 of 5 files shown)


charts/prefect-server/values.yaml
```diff
@@ -442,6 +442,26 @@ backgroundServices:
...
+    # Docket configuration for background service coordination
+    # ref: https://docs.prefect.io/v3/advanced/self-hosted#docket-url-for-background-services
+    # Docket coordinates background services like the scheduler, late run detection,
+    # and automation triggers. By default it uses in-memory storage (memory://),
+    # which only works for single-server deployments.
+    # For high-availability deployments, configure Docket to use Redis.
+    docket:
+      # -- URL for the Docket scheduler backend
+      # examples: redis://host:6379/0, redis://host:6379/0 (TLS), redis://user:pass@host:6379/0
+      # leave empty to use the default in-memory backend (memory://)
```
mitchnielsen:
```suggestion
      # -- URL for the Docket scheduler backend.
      # Examples: redis://host:6379/0, redis://host:6379/0 (TLS), redis://user:pass@host:6379/0.
      # Leave empty to use the default in-memory backend (memory://).
```

this should make the README entry more legible, because currently it's:

> name of an existing Kubernetes secret containing the Docket URL takes precedence over the url field above the secret should contain a key with the full Redis URL including any credentials

mitchnielsen (approved):

PR #618, opened 2026-04-10
feat: deploy HTTPRoute without the need of Gateway creation

### LINKED ISSUE #616: HTTP Route is only deployed, when Gateway is enabled
The HTTPRoute of the Prefect Server is only Deployed, when Gateway is enabled.
But there are valid use cases when the prefect server MUST use an existing Gateway and only a new HTTPRoute should be deployed.
[File: httproute.yaml, Line 1](https://github.com/PrefectHQ/prefect-helm/blob/2026.3.19205444/charts/prefect-server/templates/httproute.yaml#L1)
[File: gateway.yaml, Line 1](https://github.com/PrefectHQ/prefect-helm/blob/2026.3.19205444/charts/prefect-server/templates/gateway.yaml#L3)

**Current behavior**
To deploy a new HTTPRoute a new Gateway is also deployed, and additionaly deployed although an existing Gateway should be used. If an existing Gateway is used by the HTTPRoute, an additional Gateway is deployed. This causes additional deployments, components and costs, but this components are newer used.

**Expected behavior**
There shall be a way to deploy a new HTTPRoute with an existing Gateway referenced without the need to deploy an additional Gateway.

**Possbile Solution**
Remove `and .Values.gateway.enabled` from [httproute.yaml, Line 1](https://github.com/PrefectHQ/prefect-helm/blob/2026.3.19205444/charts/prefect-server/templates/httproute.yaml#L1)

I can also add an MR if this will speed the process. Thanks!

dakonr:
### Summary

Closes #616 

There shall be a way to deploy a new HTTPRoute with an existing Gateway referenced without the need to deploy an additional Gateway for each Prefect deployment.

The code adds the value gateway.create which defaults to true, so that there is no change in the helm charts behavior. If the value gateway.enabled=true and gateway.create=false only a HTTPRoute is deployed. This HTTPRoute MUST reference an existing gateway i. e. an Istio gateway in another namespace which allows HTTPRoutes from other namespaces.

This follows the design principle of the Kubernetes Gateway API and shared responsibility between Cluster Operators and Application Developers. [A high-level overview can be found here.](https://gateway-api.sigs.k8s.io)

### Requirements

### DIFF
--- a/charts/prefect-server/templates/_validation.tpl
+++ b/charts/prefect-server/templates/_validation.tpl
@@ -43,6 +43,9 @@
 {{- if and .Values.gateway.enabled .Values.ingress.enabled -}}
   {{- fail "Gateway API and Ingress are mutually exclusive. Only one can be enabled at a time. Please set either gateway.enabled=false or ingress.enabled=false." -}}
 {{- end -}}
+{{- if and .Values.httproute.enabled .Values.ingress.enabled -}}
+  {{- fail "HTTPRoute and Ingress are mutually exclusive. Only one can be enabled at a time. Please set either httproute.enabled=false or ingress.enabled=false." -}}
+{{- end -}}
 {{- if and .Values.gateway.enabled (not .Values.gateway.className) -}}
   {{- fail "gateway.className is required when gateway.enabled=true. Please specify a GatewayClass that exists in your cluster." -}}
 {{- end -}}
--- a/charts/prefect-server/templates/_validation.tpl
+++ b/charts/prefect-server/templates/_validation.tpl
@@ -43,13 +43,21 @@
 {{- if and .Values.gateway.enabled .Values.ingress.enabled -}}
   {{- fail "Gateway API and Ingress are mutually exclusive. Only one can be enabled at a time. Please set either gateway.enabled=false or ingress.enabled=false." -}}
 {{- end -}}
-{{- if and .Values.httproute.enabled (not .Values.gateway.create) (eq (len .Values.httproute.parentRefs) 0) -}}
-  {{- fail "When httproute.enabled=true and gateway.create=false, at least one entry in httproute.parentRefs is required. Set one or more parentRefs[].name values to attach the HTTPRoute to an existing Gateway." -}}
+{{- if and .Values.httproute.enabled (not .Values.gateway.create) -}}
+  {{- $hasName := false -}}
+  {{- range .Values.httproute.parentRefs -}}
+    {{- if .name -}}
+      {{- $hasName = true -}}
+    {{- end -}}
+  {{- end -}}
+  {{- if not $hasName -}}
+    {{- fail "When httproute.enabled=true and gateway.create=false, at least one entry in httproute.parentRefs is required. Set one or more parentRefs[].name values to attach the HTTPRoute to an existing Gateway." -}}
+  {{- end -}}
 {{- end -}}
-{{- if and .Values.gateway.enabled (not .Values.gateway.className) -}}
-  {{- fail "gateway.className is required when gateway.enabled=true. Please specify a GatewayClass that exists in your cluster." -}}
+{{- if and .Values.gateway.enabled .Values.gateway.create (not .Values.gateway.className) -}}
+  {{- fail "gateway.className is required when gateway.enabled=true and gateway.create=true. Please specify a GatewayClass that exists in your cluster." -}}
 {{- end -}}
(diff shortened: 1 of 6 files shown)


charts/prefect-server/templates/_validation.tpl
```diff
@@ -43,6 +43,9 @@
 {{- if and .Values.gateway.enabled .Values.ingress.enabled -}}
   {{- fail "Gateway API and Ingress are mutually exclusive. Only one can be enabled at a time. Please set either gateway.enabled=false or ingress.enabled=false." -}}
 {{- end -}}
+{{- if and .Values.httproute.enabled (not .Values.gateway.create) (eq (len .Values.httproute.parentRefs) 0) -}}
+  {{- fail "When httproute.enabled=true and gateway.create=false, at least one entry in httproute.parentRefs is required. Set one or more parentRefs[].name values to attach the HTTPRoute to an existing Gateway." -}}
+{{- end -}}
 {{- if and .Values.gateway.enabled (not .Values.gateway.className) -}}
```
mitchnielsen:
```suggestion
if and .Values.gateway.enabled .Values.gateway.create (not .Values.gateway.className)
```

The existing check if and .Values.gateway.enabled (not .Values.gateway.className) will fail the template if className is empty, even when no Gateway is being created. The test only passes because it sets className: "istio". A user setting gateway.create=false should not be forced to provide a className that won't be used.

charts/prefect-server/templates/_validation.tpl
mitchnielsen:
```suggestion
{{- if and .Values.gateway.enabled .Values.gateway.create (not (include "gateway.apiAvailable" .)) -}}
```

The gateway.apiAvailable check requires the Gateway CRD to exist. When gateway.create=false, the cluster may only have the HTTPRoute CRD (the Gateway CRD is only needed on the cluster where the Gateway controller runs).

charts/prefect-server/templates/gateway.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.gateway.enabled }}
+{{- if and .Values.gateway.enabled .Values.gateway.create}}
```
mitchnielsen:
```suggestion
{{- if and .Values.gateway.enabled .Values.gateway.create }}
```

charts/prefect-server/templates/_validation.tpl
```diff
@@ -43,6 +43,9 @@
 {{- if and .Values.gateway.enabled .Values.ingress.enabled -}}
   {{- fail "Gateway API and Ingress are mutually exclusive. Only one can be enabled at a time. Please set either gateway.enabled=false or ingress.enabled=false." -}}
 {{- end -}}
+{{- if and .Values.httproute.enabled (not .Values.gateway.create) (eq (len .Values.httproute.parentRefs) 0) -}}
```
mitchnielsen: `len .Values.httproute.parentRefs` checks httproute.parentRefs length, but default value has one entry. The default parentRefs in values.yaml has one entry with name: "". So `(eq (len .Values.httproute.parentRefs) 0)` will never be true with default values.

The intent is to catch users who don't provide a named parentRef, but the check only catches a completely empty list. It should probably also check that at least one entry has a non-empty name.

Otherwise the user gets a rendered HTTPRoute with `parentRefs[0].name: ""` rather than a helpful error.

charts/prefect-server/tests/gateway_test.yaml
```diff
@@ -185,6 +185,28 @@ tests:
         isKind:
           of: HTTPRoute
 
+  - it: Should create only HTTPRoute when Gateway creation is disabled
```
mitchnielsen, line 188: Thanks for adding this test case. Let's add some more to confirm the validation logic catching config errors works as expected.

mitchnielsen (changes requested):

charts/prefect-server/tests/gateway_test.yaml
dakonr: Hi @mitchnielsen 
thanks for your feedback. I implemented all annotations and recommendations that you have.
I also added a more specific validation and more test cases.
Just for transparency reasons: some of the test cases are generated but seems legit to me and will pass.
I'm looking forward to get feedback to my resent changes soon! Thanks!

mitchnielsen (approved):

mitchnielsen:
Thanks again @dakonr!

PR #620, opened 2026-04-21
fix(prefect-server): dedupe env vars on migration Job

### LINKED ISSUE #602: No possibility to set env var on server to different value on migration job
The migration job gets all environment variables from `global.prefect.env`, `server.env` and `migrations.env`. I would however like to set `PREFECT_API_DATABASE_CONNECTION_TIMEOUT` to a higher value on the migrations than on the server, because of timeout issues on the migrations job. This is now impossible, since specifying the env var in both sections with different values, results in duplicate entries of the environment variable in the Job spec. 

Minimal reproduction values: 
```values.yaml
server:
  env:
    - name: PREFECT_API_DATABASE_TIMEOUT
      value: "30"
backgroundServices:
  runAsSeparateDeployment: true
migrations:
  enabled: true
  env:
    - name: PREFECT_API_DATABASE_TIMEOUT
      value: "120"
redis:
  enabled: true
```

mitchnielsen:
### Summary

The migration Job concatenated env entries from `global.prefect.env`, `server.env`, and `migrations.env`, so setting the same variable in more than one source (e.g. a longer `PREFECT_API_DATABASE_TIMEOUT` for migrations only) produced duplicate entries in the Job spec. Kubernetes tolerated it, but validators and GitOps diffs flagged it as broken.

Merged the three sources in a helper with precedence `migrations.env` > `server.env` > `global.prefect.env`. Kept the fix scoped to the migration Job; the same concat pattern in `deployment.yaml` and `background-services/deployment.yaml` was left alone since the issue is specific to migrations and the blast radius on those templates is larger.

README churn beyond the `migrations.env` row is from running the pre-commit `helm-docs` hook, which caught existing drift between `values.yaml` comments and the generated README.

Closes #602

### Requirements

- [x] Body includes \`Closes <issue>\`, if available
- [ ] Deprecation/removal checks are added in \`templates/NOTES.txt\`
- [x] \`Draft\` status is used until ready for review

<details>
<summary>Session context</summary>

Walked through the issue repro in #602, confirmed the root cause was plain list concatenation in `templates/pre-upgrade-hook.yaml`. Considered a dict-based dedupe but went with reverse-iteration-and-skip so output stays a plain list and no Sprig feature beyond `hasKey`/`set` is needed. Verified via `helm template` against the issue's exact values and via a new set of helm-unittest cases covering each pairwise and three-way override.

</details>

### DIFF
--- a/charts/prefect-server/templates/_helpers.tpl
+++ b/charts/prefect-server/templates/_helpers.tpl
@@ -106,6 +106,39 @@ Priority:
 {{- end }}
 {{- end -}}
 
+{{- /*
+    server.migrations.envVars:
+      Merged env vars for the migrations Job, deduplicated by name so the
+      same variable is never emitted twice in the Job spec.
+
+      Precedence (highest wins):
+        1. .Values.migrations.env
+        2. .Values.server.env
+        3. .Values.global.prefect.env
+
+      Iterates highest-precedence first and appends each entry only if its
+      name has not yet been seen, so the emitted list has unique names with
+      the highest-precedence value for each. Order within the output is
+      highest-precedence source first, which is semantically equivalent in
+      Kubernetes (env order does not affect resolution).
+*/ -}}
+{{- define "server.migrations.envVars" -}}
+{{- $sources := list .Values.migrations.env .Values.server.env .Values.global.prefect.env -}}
+{{- $seen := dict -}}
+{{- $merged := list -}}
+{{- range $source := $sources -}}
+  {{- range $entry := $source -}}
+    {{- if not (hasKey $seen $entry.name) -}}
+      {{- $_ := set $seen $entry.name true -}}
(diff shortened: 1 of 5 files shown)


charts/prefect-server/templates/_helpers.tpl
```diff
@@ -106,6 +106,39 @@ Priority:
 {{- end }}
 {{- end -}}
 
+{{- /*
```
jimid27, line 109: Just verified you have tested this locally with helm template?

jimid27 (approved):
lgtm just a question

charts/prefect-server/templates/_helpers.tpl
mitchnielsen: Yep! Plus that's what the helm unit tests do

PR #630, opened 2026-06-10
feat(prefect-server): support database URL components

### LINKED ISSUE #528: Allow for manually setting Database connection components from existing secrets
As an administrator, I want to be able to set the database connection string from components:

- `PREFECT_SERVER_DATABASE_DRIVER` (e.g. `postgresql+asyncpg`)
- `PREFECT_SERVER_DATABASE_HOST`
- `PREFECT_SERVER_DATABASE_PORT`
- `PREFECT_SERVER_DATABASE_USER`
- `PREFECT_SERVER_DATABASE_PASSWORD`
- `PREFECT_SERVER_DATABASE_NAME`

I should be able to do this when I set:

```
secret:
  create: false
server:
  extraEnvVarsSecret: prefect-db-secret
postgresql:
  enabled: false
```

Today, this will still add the `PREFECT_API_DATABASE_CONNECTION_URL` env, which will be unusable because it doesn't exist in this setup.

mitchnielsen:
### Summary

Some external database operators create immutable credential Secrets with only component keys like `username` and `password`. This adds a `database` values block that can compose the Prefect database connection URL from mixed literal values and Secret keys, including for the migration Job.

Closes #528

Specific context: https://github.com/PrefectHQ/prefect-helm/issues/528#issuecomment-4669272200

### Requirements

- [x] Body includes `Closes <issue>`, if available (using `Related to #528` per PR convention)
- [x] Deprecation/removal checks are added in `templates/NOTES.txt` (not applicable; no deprecation/removal)

<details>
<summary>Session context</summary>

The request came from users relying on the Zalando Postgres Operator, which creates credential Secrets that cannot be modified to include a full connection string. The existing `extraEnvVarsSecret` path helps deployments but does not cover the migration Job, so this keeps the database env rendering shared across server, background services, and migrations.

I kept the existing `secret.*` and bundled PostgreSQL behavior as the default path, with `database.connectionString` and `database.connectionStringSecretKey` taking precedence when the new block is used.
</details>

### DIFF
--- a/charts/prefect-server/README.md
+++ b/charts/prefect-server/README.md
@@ -398,7 +398,7 @@ the HorizontalPodAutoscaler.
 | commonLabels | object | `{}` | labels to add to all deployed objects |
 | externalDatabase.connectionString | string | `""` | literal database connection URL. Takes precedence over component settings |
 | externalDatabase.connectionStringSecretKey | string | `""` | key in `externalDatabase.existingSecret` containing the full database connection URL |
-| externalDatabase.driver | string | `"postgresql+asyncpg"` | database driver used when composing a connection URL from components |
+| externalDatabase.driver | string | `""` | database driver used when composing a connection URL from components (e.g. `postgresql+asyncpg`) |
 | externalDatabase.driverSecretKey | string | `""` | key in `externalDatabase.existingSecret` containing the database driver |
 | externalDatabase.existingSecret | string | `""` | name of an existing Secret to read database settings from |
 | externalDatabase.host | string | `""` | database host used when composing a connection URL from components |
@@ -407,7 +407,7 @@ the HorizontalPodAutoscaler.
 | externalDatabase.nameSecretKey | string | `""` | key in `externalDatabase.existingSecret` containing the database name |
 | externalDatabase.password | string | `""` | database password used when composing a connection URL from components |
 | externalDatabase.passwordSecretKey | string | `""` | key in `externalDatabase.existingSecret` containing the database password |
-| externalDatabase.port | string | `"5432"` | database port used when composing a connection URL from components |
+| externalDatabase.port | string | `""` | database port used when composing a connection URL from components (e.g. `5432`) |
 | externalDatabase.portSecretKey | string | `""` | key in `externalDatabase.existingSecret` containing the database port |
 | externalDatabase.username | string | `""` | database username used when composing a connection URL from components |
 | externalDatabase.usernameSecretKey | string | `""` | key in `externalDatabase.existingSecret` containing the database username |
--- a/charts/prefect-server/README.md
+++ b/charts/prefect-server/README.md
@@ -185,9 +185,7 @@ Two secrets are created when not providing an existing secret name:
 
 If you want to disable the bundled PostgreSQL chart and use an external instance, provide the following configuration:
 
-The `externalDatabase` block is only for external databases. For databases managed by this chart, continue to configure the bundled PostgreSQL chart with `postgresql` or the embedded SQLite database with `sqlite`.
-
-Database connection configuration is applied in this order:
+The `externalDatabase` block is only for external databases. For databases managed by this chart, continue to configure the bundled PostgreSQL chart with `postgresql` or the embedded SQLite database with `sqlite`. Database connection configuration is applied in this order:
 
 1. `sqlite.enabled=true` uses the embedded SQLite database.
 2. `externalDatabase.connectionString` uses a literal external database connection URL.
--- a/charts/prefect-server/README.md
+++ b/charts/prefect-server/README.md
@@ -185,6 +185,16 @@ Two secrets are created when not providing an existing secret name:
 
 If you want to disable the bundled PostgreSQL chart and use an external instance, provide the following configuration:
 
+The `database` block is only for external databases. For databases managed by this chart, continue to configure the bundled PostgreSQL chart with `postgresql` or the embedded SQLite database with `sqlite`.
+
+Database connection configuration is applied in this order:
+
+1. `sqlite.enabled=true` uses the embedded SQLite database.
+2. `database.connectionString` uses a literal external database connection URL.
+3. `database.existingSecret` with `database.connectionStringSecretKey` reads a full external database connection URL from an existing Secret.
+4. `database` component fields and Secret keys compose an external database connection URL from driver, username, password, host, port, and name.
+5. Existing `secret`/`postgresql` behavior is used when none of the above `database` options are set.
+
 ```yaml
 postgresql:
   enabled: false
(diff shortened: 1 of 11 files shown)


ArcticXWolf:
Looks great, thanks for implementing this ❤️ 

Just one comment towards documentation: Maybe we should mention that the new `database` block is only to configure an external database, for internal ones you would continue to use the `postgresql` and `sqlite` blocks. And maybe we should also mention what the precedence rules between the three blocks are.
Both those infos should probably be in the README.md and also in the values.yaml comments.

mitchnielsen:
My pleasure, thanks for the detailed suggestions and feedback 🤝 

I ended up renaming the key `externalDatabase` to make this more clear. How's it looking now overall?

charts/prefect-server/templates/_helpers.tpl
```diff
@@ -246,6 +246,111 @@ Priority:
 {{- end -}}
 {{- end -}}
 
+{{/*
+  server.database-has-component-config:
+    Returns true when the new database component configuration is in use.
+*/}}
+{{- define "server.database-has-component-config" -}}
+{{- $db := .Values.externalDatabase | default dict -}}
+{{- if or $db.username $db.usernameSecretKey $db.password $db.passwordSecretKey $db.host $db.hostSecretKey $db.name $db.nameSecretKey $db.driverSecretKey $db.portSecretKey -}}
```
ArcticXWolf: I think here the $db.driver and $db.port is missing.

charts/prefect-server/tests/database_test.yaml
```diff
@@ -180,3 +180,132 @@ tests:
...
+                key: connection-string
+      - template: deployment.yaml
+        notContains:
+          path: .spec.template.spec.containers[0].env
+          content:
+            name: PREFECT_SERVER_DATABASE_USER
+            valueFrom:
+              secretKeyRef:
+                name: database-connection
+                key: username
```
ArcticXWolf, line 311: Not sure how expansive the test suite should be, but I think we should add these two tests in case someone messes up a single specific database option expansion (e.g. port or such):

* one test where we set everything via literal values (so all literal values work)
* one test where we set everything via secret (so all secret expansion work)

charts/prefect-server/README.md
```diff
@@ -185,6 +185,16 @@ Two secrets are created when not providing an existing secret name:
 
 If you want to disable the bundled PostgreSQL chart and use an external instance, provide the following configuration:
 
+The `externalDatabase` block is only for external databases. For databases managed by this chart, continue to configure the bundled PostgreSQL chart with `postgresql` or the embedded SQLite database with `sqlite`.
+
```
ArcticXWolf: Maybe we can merge these two paragraphs (especially since the first paragraph ends on a colon, but isnt followed by a configuration example)

charts/prefect-server/values.schema.json
```diff
@@ -1423,6 +1423,31 @@
...
+        "usernameSecretKey": { "type": "string" },
+        "password": { "type": "string" },
+        "passwordSecretKey": { "type": "string" },
+        "host": { "type": "string" },
+        "hostSecretKey": { "type": "string" },
+        "port": {
+          "type": [
+            "integer",
+            "string"
+          ]
```
ArcticXWolf, line 1444: The tests (and documentation) only specify port as a string. As far as I can see, you've added a ` | quote` in most places, but I am unsure if the line 297 in `_helper.tpl` needs one too.

Just using the port as an integer in one of the tests would probably confirm this anyway.

charts/prefect-server/values.yaml
```diff
@@ -707,6 +707,42 @@ httproute:
...
+  # -- key in `externalDatabase.existingSecret` containing the database host
+  hostSecretKey: ""
+  # -- database port used when composing a connection URL from components
+  port: "5432"
+  # -- key in `externalDatabase.existingSecret` containing the database port
+  portSecretKey: ""
+  # -- database name used when composing a connection URL from components
+  name: ""
+  # -- key in `externalDatabase.existingSecret` containing the database name
+  nameSecretKey: ""
```
ArcticXWolf, line 744: Maybe we should remove the default values for driver and port and instead put them as examples in the comment.

ArcticXWolf (commented):
LGTM, great work! Thank you for that!

I only added some cosmetic comments.

Also I didnt test this against a real cluster.

charts/prefect-server/templates/_helpers.tpl
ArcticXWolf: I just noticed that they are probably missing because we have default values for them in the `values.yaml`.

I still think that removing the default values, adding them to the comment and then adding port and driver to this check is the better way, but maybe you had a specific reason for that.

mitchnielsen: Good catch, and you're right about why they were missing. Went with your suggestion in 38413ff: removed the driver and port defaults from values.yaml (now examples in the comments) and added both to the database-has-component-config check.

The postgresql+asyncpg / 5432 fallbacks still apply at render time in _helpers.tpl, so a partially-configured block still gets sensible defaults. They just no longer live in values.yaml where an unset block would falsely look configured.

charts/prefect-server/values.yaml
mitchnielsen: Done in 38413ff — driver and port now default to "", with postgresql+asyncpg and 5432 shown as examples in the comments.

charts/prefect-server/values.schema.json
mitchnielsen: Confirmed in 1a140ee by adding a test that passes port: 5432 as an integer. It renders as PREFECT_SERVER_DATABASE_PORT: "5432" and composes into the URL as ...:5432/... correctly, so no additional quoting was needed.

charts/prefect-server/tests/database_test.yaml
mitchnielsen: Added both in 1a140ee — one setting every component as a literal value (with the integer port above), one setting every component via a Secret key. Each asserts the individual env vars plus the composed connection URL, so a single mis-wired component now fails the suite.

charts/prefect-server/README.md
mitchnielsen: Merged the paragraphs in 75b3a3e, then in 39a1a90 reworked the intro so the dangling colon is actually gone. The lead-in sentence now introduces the externalDatabase block and ends on a period. The only remaining colon is on "applied in this order:", which is immediately followed by the precedence list.

mitchnielsen:
Really appreciate the feedback @ArcticXWolf 🤝

ArcticXWolf:
Changes are looking good!
I cannot approve the PR since I do not have the permissions, but it is approved from me 😉

jamiezieziula (approved):

PR #652, opened 2026-08-04
prefect-server: land PostgreSQL subchart bump 12.x -> 18.x safely

jamiezieziula:
Lands the Bitnami PostgreSQL subchart major bump (`12.12.10` -> `18.8.5`) safely, as the follow-up to the DO-NOT-MERGE dependency PR #650.

**Approach:** take the subchart to 18.x for the current template/values/security machinery, but hold the bundled engine on the 14.x line via the existing `bitnamilegacy` image override. This avoids the two things that made the raw bump unsafe: the destructive pg 15->17 migration on the default path, and the non-pullable `bitnami/postgresql:latest` default. Mirrors the existing redis subchart precedent.

## Changes
- `Chart.yaml`: postgresql `12.12.10` -> `18.8.5`
- `values.yaml`: carry the image override forward (`registry`/`repository`/`tag`), pinned to `bitnamilegacy/postgresql:14.18.0` (in-place patch bump from `14.13.0`, the final pg14 patch release — picks up security fixes, no data migration)
- `UPGRADING.md`: no-action-by-default note, the tightened `securityContext` defaults table, and an opt-in destructive pg 15->17 runbook
- `README.md`: regenerated with `helm-docs 1.14.2`

## Validation
- 217/217 helm unit tests pass (pinned `helmunittest/helm-unittest:3.18.4-1.0.0`)
- `helm template` verified: clean install renders `docker.io/bitnamilegacy/postgresql:14.18.0` (manifest confirmed pullable); upgrade-from-14.x preserves the PVC; container `securityContext` is `restricted`-PSA compliant; `postgresql.enabled=false` path unaffected

## Follow-up
- `bitnamilegacy` is a frozen archive (no new builds), so `14.18.0` is the last pullable pinned pg14 image. A future engine upgrade to pg 17/18 or a move to a maintained image source is tracked separately.

## Remaining (needs a cluster, pre-merge gate)
- Live `helm upgrade` against a real pre-existing 14.x PVC (pod starts, data survives)
- Live clean install: pod start + prefect-server connection + migration job

Resolves PLA-3010


mitchnielsen:
With all the included testing/validation this sounds safe to me, I just hesitate to let the chart version drift that far from the application version.

It looks like it includes some security-related changes, which we could (maybe) already configure?

Thinking out loud, I'd probably lean to skipping this for now and then doing a true PG 18 upgrade (chart + image tag) together in one go.

Any strong feelings?

PR #680, opened 2026-08-18
feat(prefect-server): allow disabling service account token automount

sinacek:
### Summary

  The `prefect-server` chart always mounts a service account API token into the
  server, background-services and pre-upgrade-hook pods, and there is no value to
  turn it off. Prefect Server does not call the Kubernetes API — only the worker
  does, to create flow run jobs — so the token is a credential sitting in a pod
  that has no use for it. Clusters that enforce CIS Kubernetes Benchmark 5.1.6
  ("ensure that Service Account Tokens are only mounted where necessary") flag
  this, and today the only workaround is patching the `default` ServiceAccount
  outside of Helm.

  This adds `automountServiceAccountToken` to:

  - `server` and `backgroundServices` pod specs (`deployment.yaml`,
    `background-services/deployment.yaml`, `pre-upgrade-hook.yaml`)
  - both ServiceAccount templates (`serviceaccount.yaml`,
    `background-services/serviceaccount.yaml`)

  Both pod-level and ServiceAccount-level knobs are needed: the SA-level one has
  no effect when `serviceAccount.create: false` and the deployment runs under an
  externally managed service account.

  All four values default to `true`, so rendered output is unchanged unless a user
  opts in. The bundled `postgresql` and `redis` subcharts already expose the same
  setting, so this brings the chart's own resources in line.

  Verified locally: `helm unittest` (220 tests, incl. 2 new ones covering the
  default and the disabled case), `ct lint`, `helm-docs` regeneration (README diff
  is exactly the four new value rows), and `yamllint --strict`.

  ### Requirements

  - [ ] Body includes `Closes <issue>`, if available
  - [ ] Relevant labels are added
  - [ ] `Draft` status is used until ready for review

### DIFF
--- a/charts/prefect-server/README.md
+++ b/charts/prefect-server/README.md
@@ -377,6 +377,7 @@ the HorizontalPodAutoscaler.
 |-----|------|---------|-------------|
 | backgroundServices.affinity | object | `{}` | affinity for background-services pod assignment |
 | backgroundServices.args | list | `[]` | Custom container command arguments |
+| backgroundServices.automountServiceAccountToken | bool | `true` | whether the background services pod mounts a service account API token. background services do not call the kubernetes api, so this can be set to false |
 | backgroundServices.command | list | `[]` | Custom container entrypoint |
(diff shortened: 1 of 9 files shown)


jamiezieziula (approved):
Thanks!

